PR #86, opened 2022-02-16
Create reporter per command run to avoid write issues

egorio:
The `NO autofix` test case was not in the right place and never run.

Another issues came out of blue after I fixed the test:

```
 Error: Error: write after end
    at DerivedLogger.<anonymous> (/Users/everbits/Sites/rehearsal-js/packages/reporter/src/reporter.ts:181:13)
    at DerivedLogger.emit (node:events:390:28)
    at DerivedLogger.emit (node:domain:475:12)
    at errorOrDestroy (/Users/everbits/Sites/rehearsal-js/node_modules/readable-stream/lib/internal/streams/destroy.js:98:101)
    at writeAfterEnd (/Users/everbits/Sites/rehearsal-js/node_modules/readable-stream/lib/_stream_writable.js:263:3)
    at DerivedLogger.Writable.write (/Users/everbits/Sites/rehearsal-js/node_modules/readable-stream/lib/_stream_writable.js:305:21)
    at DerivedLogger.log (/Users/everbits/Sites/rehearsal-js/node_modules/winston/lib/winston/logger.js:252:14)
    at Object.run (/Users/everbits/Sites/rehearsal-js/packages/plugin-ts-migrate/src/plugin-ts-migrate-autofix.ts:137:25)
```

To avoid this we need to create a separate reporter object for every command run.

### DIFF
--- a/packages/cli/src/commands/ts.ts
+++ b/packages/cli/src/commands/ts.ts
@@ -33,9 +33,6 @@ const DEFAULT_TS_BUILD = "beta";
 let TSC_PATH = "";
 let TS_MIGRATE_PATH = "";
 
-// cwd for development only
-const REPORTER = new Reporter({ cwd: process.cwd() });
-
 type Context = {
   tsVersion: string;
   latestELRdBuild: string;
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/ts.ts
```diff
@@ -33,9 +33,6 @@ const DEFAULT_TS_BUILD = "beta";
 let TSC_PATH = "";
 let TS_MIGRATE_PATH = "";
 
-// cwd for development only
```
chadhietala, line 36: @lynchbomb why was this here in the first place?

chadhietala (commented):
Just a quick question for @lynchbomb

packages/cli/src/commands/ts.ts
lynchbomb: So we can inspect the file report during dev at various states before its compiled output. We use tmp dir so those steps are lost otherwise.

lynchbomb (approved):

PR #122, opened 2022-03-14
refactor: simplify .gitignore

egorio:


.gitignore
```diff
@@ -1,18 +1,15 @@
-.Rapp.history
-dist/
-node_modules/
-vendor/
-tmp/
-/*workspace
+.DS_Store
+.rehearsal.json
```
lynchbomb, line 2: we have mocks where we need to modify checkin this file. which is the exception paths we listed

.gitignore
```diff
@@ -1,18 +1,15 @@
...
+.nyc_output/
+.vscode/
+
 *.log
 *.tgz
 *.tmp
-.nyc_output
-.idea/*
-packages/cli/bin/run.js
-**SCRATCH**
```
lynchbomb, line 13: still needed

egorio: Where this `**SCRATCH**` came from?

PR #128, opened 2022-03-25
refactor: init migration package

egorio:
NOT READY


packages/migration/src/transforms/6133-fix-transform.ts
```diff
@@ -0,0 +1,21 @@
+import ts from 'typescript';
+
+import FixTransform from '../interfaces/fix-transform';
+import { transformDiagnosedNode } from '../helpers/typescript-ast';
+
+export default class FixTransform6133 extends FixTransform {
```
egorio, line 6: This Fix, as an example, currently covers 2 cases: unused import declaration and unused import specifier (function name)

PR #156, opened 2022-04-15
fix: missing tests files and remove used dependencies and variables

egorio:


packages/cli/test/commands/ts.test.ts
```diff
@@ -130,8 +130,9 @@ describe('ts:command tsc version check', async () => {
       '--is_test',
     ]);
 
+    // TODO: Fix CLI or this test
     expect(ctx.stdout).to.contain(
-      `This application is already on the latest version of TypeScript@${TEST_TSC_VERSION}`
```
egorio, line 134: Seems CLI works not as we expect :( Will fix later.

chadhietala (approved):

PR #157, opened 2022-04-22
feat: add placeholder transformer for TS2322 error code 

brendaHuang:
Add the transformer for 2322. We may add more logic later. Also revise the test.

### DIFF
--- a/packages/migrate/src/transforms/2322-fix-transform.ts
+++ b/packages/migrate/src/transforms/2322-fix-transform.ts
@@ -1,0 +1,5 @@
+import FixTransform from '../interfaces/fix-transform';
+
+export default class FixTransform2322 extends FixTransform {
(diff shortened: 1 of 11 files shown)


packages/migrate/src/transforms/2322-fix-transform.ts
```diff
@@ -0,0 +1,5 @@
+import FixTransform from '../interfaces/fix-transform';
+
+export default class FixTransform2322 extends FixTransform {
+  hint = `Type '{0}' is not assignable to type '{1}'.`;
```
egorio, line 4: Please, add useful hint instead default error message

brendaHuang: done

egorio (approved):

PR #176, opened 2022-05-19
feat: Allow multiple file updates per plugin call

brendaHuang:
This PR allows each plugin to update multiple files.

### DIFF
--- a/packages/migrate/src/interfaces/fix-transform.ts
+++ b/packages/migrate/src/interfaces/fix-transform.ts
@@ -1,13 +1,17 @@
 import ts from 'typescript';
 
+export interface FixedFile {
+  fileName: string;
+  text: string;
+}
 export default class FixTransform {
   /** Engineer friendly message describes steps needs to be done to fix the related issue */
   hint?: string;
 
   /** Function to fix the diagnostic issue */
-  fix?: (diagnostic: ts.DiagnosticWithLocation, service: ts.LanguageService) => string;
+  fix?: (diagnostic: ts.DiagnosticWithLocation, service: ts.LanguageService) => FixedFile[];
 
-  run(diagnostic: ts.DiagnosticWithLocation, service: ts.LanguageService): string {
-    return this.fix ? this.fix(diagnostic, service) : diagnostic.file.getFullText();
+  run(diagnostic: ts.DiagnosticWithLocation, service: ts.LanguageService): FixedFile[] {
+    return this.fix ? this.fix(diagnostic, service) : [];
   }
 }
--- a/packages/migrate/src/interfaces/plugin.ts
+++ b/packages/migrate/src/interfaces/plugin.ts
@@ -9,7 +9,7 @@ export type PluginParams<PluginOptions> = {
   options?: PluginOptions;
 };
 
-export type PluginResult = Promise<string>;
+export type PluginResult = Promise<string[]>;
 
 export class Plugin {
   protected readonly logger?: winston.Logger;
@@ -23,6 +23,6 @@ export class Plugin {
   }
 
   async run(params: PluginParams<PluginOptions>): PluginResult {
-    return this.service.getFileText(params.fileName);
+    return [params.fileName];
   }
 }
--- a/packages/migrate/src/migrate.ts
+++ b/packages/migrate/src/migrate.ts
@@ -62,14 +62,16 @@ export default async function migrate(input: MigrateInput): Promise<MigrateOutpu
   for (const fileName of fileNames) {
     logger?.info(`Processing file: ${fileName}`);
 
+    let allChangedFiles: Set<string> = new Set();
+
     for (const pluginClass of plugins) {
       const plugin = new pluginClass(service, logger, reporter);
-      const updatedText = await plugin.run({ fileName });
-      service.setFileText(fileName, updatedText);
+      const changedFiles = await plugin.run({ fileName });
+      allChangedFiles = new Set([...allChangedFiles, ...changedFiles]);
     }
 
     // Save file to the filesystem
-    service.saveFile(fileName);
+    allChangedFiles.forEach((file) => service.saveFile(file));
   }
 
   logger?.info(`Migration finished.`);
--- a/packages/migrate/src/plugins/diagnostics-autofix.plugin.ts
+++ b/packages/migrate/src/plugins/diagnostics-autofix.plugin.ts
@@ -17,26 +17,35 @@ export default class DiagnosticAutofixPlugin extends Plugin {
 
     this.logger?.debug(`Plugin 'DiagnosticAutofix' run on ${params.fileName}`);
 
+    const allFixedFiles: Set<string> = new Set();
+
     while (diagnostics.length > 0 && tries-- > 0) {
       const diagnostic = diagnostics.shift()!;
 
       const fix = getFixForDiagnostic(diagnostic);
       const node = findNodeAtPosition(diagnostic.file, diagnostic.start, diagnostic.length);
 
-      let text = fix.run(diagnostic, this.service.getLanguageService());
+      const fixedFiles = fix.run(diagnostic, this.service.getLanguageService());
 
-      const fixed = this.isSourceCodeChanged(diagnostic.file.getFullText(), text);
       const hint = this.prepareHint(diagnostic.messageText, fix?.hint);
+
+      const fixed = fixedFiles.length > 0;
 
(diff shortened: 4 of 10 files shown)


packages/migrate/src/interfaces/plugin.ts
```diff
@@ -23,6 +23,6 @@ export class Plugin {
   }
 
   async run(params: PluginParams<PluginOptions>): PluginResult {
-    return this.service.getFileText(params.fileName);
+    return [params.fileName];
```
lynchbomb, line 26: Confirming this no longer needs to be passed through the service?

packages/migrate/src/migrate.ts
```diff
@@ -62,14 +62,16 @@ export default async function migrate(input: MigrateInput): Promise<MigrateOutpu
   for (const fileName of fileNames) {
     logger?.info(`Processing file: ${fileName}`);
 
+    let allChangedFiles: Set<string> = new Set();
+
     for (const pluginClass of plugins) {
       const plugin = new pluginClass(service, logger, reporter);
-      const updatedText = await plugin.run({ fileName });
-      service.setFileText(fileName, updatedText);
+      const changedFiles = await plugin.run({ fileName });
```
lynchbomb, line 69: Again here please confirm this no longer needs to be passed though the service

packages/migrate/src/plugins/lint.plugin.ts
```diff
@@ -7,23 +7,24 @@ import { Plugin, PluginParams, PluginResult } from '../interfaces/plugin';
...
+        return [fileName];
       }
     } catch (e) {
       this.logger?.error(`Plugin 'Lint' failed on ${params.fileName}: ${(e as Error).message}`);
     }
 
     this.logger?.debug(`Plugin 'Lint' run on ${params.fileName} with no changes`);
 
-    return text;
+    return [];
```
lynchbomb, line 28: Confirm we no longer need to return the text

packages/migrate/src/transforms/2322-fix-transform.ts
```diff
@@ -1,23 +1,22 @@
...
         this.hint = `The function expects to return '{1}', but '{0}' is returned. Please convert '{0}' value to '{1}' or update the function's return type.`;
       } else if (ts.isIdentifier(node)) {
         this.hint = `The variable '${node.escapedText}' has type '{1}', but '{0}' is assigned. Please convert '{0}' to '{1}' or change variable's type.`;
       }
-
       return node;
     });
 
-    return diagnostic.file.getFullText();
+    return [];
```
lynchbomb, line 20: Confirm we no longer need to return the text

lynchbomb (changes requested):
Looking good. A few suggested changes and confirms

packages/migrate/src/interfaces/fix-transform.ts
```diff
@@ -1,13 +1,19 @@
 import ts from 'typescript';
 
+interface FixedFile {
+  fileName: string;
+  text: string;
+}
+
+export type FixedFiles = FixedFile[];
```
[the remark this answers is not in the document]
brendaHuang: done

packages/migrate/src/interfaces/plugin.ts
brendaHuang: Plugin input: changed file from transformers, including file name and new text
What Plugin does: sets the new text with the service, and iterates the changed text through diagnostics, setting the again changed text
Plugin output: **an array of names of files that have been changed, or an empty array if none has changed**
migrate.ts input: **an array of names of files that have been changed, or an empty array if none has changed**
What migrate.ts does: save each file with their final text with the service

packages/migrate/src/migrate.ts
brendaHuang: same as above

packages/migrate/src/plugins/lint.plugin.ts
brendaHuang: If there's changed text, we use setFileText() to commit the text to this file in memory and we only return the filename. If no file has changed, we return an empty array.

packages/migrate/src/plugins/empty-lines-preserve.plugin.ts
```diff
@@ -1,15 +1,17 @@
...
-    const text = this.service.getFileText(params.fileName);
+    const { fileName } = params;
+    const text = this.service.getFileText(fileName);
 
-    this.logger?.debug(`Plugin 'EmptyLinesPreserve' run on ${params.fileName}`);
+    this.logger?.debug(`Plugin 'EmptyLinesPreserve' run on ${fileName}`);
 
-    return text.replace(/^( |\t)*$/gm, `/*:line:*/`);
+    const newText = text.replace(/^( |\t)*$/gm, `/*:line:*/`);
+    this.service.setFileText(fileName, newText);
```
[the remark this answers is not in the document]
egorio: Probably not... still thinking :)

packages/migrate/src/interfaces/plugin.ts
egorio: After this PR:
**cli** - **validates user input**, pre-processes package.json and post-process **report / PR** 
↓
**migrate** - configures list of plugins, files to apply plugins and compiler options, then **run a service**
↓
**service** - compiles program and provides access to source files (cached files)... runs plugins on files (_to be done_) and saves updated files to a file system (sync) after all plugins applied to file.
↓
**plugin** - **receive filename to apply changes**, uses service to access file content... can also access to different files of create new ones, **returns the list of files been touched** (after this PR)

**Short:**
- cli - user input and output
- migrate - setup the project to compile
- service - access to source files in file system and logic to run plugins
- plugin - responsible to modify one or more files (cached)
- transform - part of the one plugin, responsible to modify AST

Seems legit :)

packages/migrate/src/plugins/empty-lines-preserve.plugin.ts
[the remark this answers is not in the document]
egorio: Let's call this.service.setFileText for now, [seems OK](https://github.com/rehearsal-js/rehearsal-js/pull/176#discussion_r879202159). We can re-think this later.

packages/migrate/src/plugins/diagnostics-autofix.plugin.ts
```diff
@@ -17,42 +17,43 @@ export default class DiagnosticAutofixPlugin extends Plugin {
 
     this.logger?.debug(`Plugin 'DiagnosticAutofix' run on ${params.fileName}`);
 
+    const allFixedFiles: Set<string> = new Set();
```
egorio, line 20: You can just use array directly:
```
const allFixedFiles: string[] = [];
...
allFixedFiles.push(fixedFile.fileName);
...
return allFixedFiles;
```

packages/migrate/src/migrate.ts
```diff
@@ -62,14 +62,16 @@ export default async function migrate(input: MigrateInput): Promise<MigrateOutpu
...
+      allChangedFiles = new Set([...allChangedFiles, ...changedFiles]);
```
egorio, line 70: The same [as below](https://github.com/rehearsal-js/rehearsal-js/pull/176/files#r879769431), you can just use `string[]` instead of `Set`:
```
allChangedFiles = [...allChangedFiles, ...changedFiles];
```

brendaHuang: The benefit of using Set is that Set will automatically merge (deduplicate) same file names. The same files can be modified by the different diagnostics. If I use an array I need to check if there are file names added twice or more times into the array.

packages/migrate/src/plugins/diagnostics-autofix.plugin.ts
brendaHuang: Use Set to deduplicate possible duplicate names of the same files that may have been changed by different diagnostics.

egorio: 👍

packages/migrate/src/migrate.ts
egorio: 👍

packages/migrate/src/interfaces/fix-transform.ts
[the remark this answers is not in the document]
brendaHuang: It has been fixed.

egorio (approved):

PR #177, opened 2022-05-25
fix: use correct comments' style inside JSX text, modules resolutions in tests

egorio:
- use correct comments' style inside JSX text
- fix global modules resolutions in tests
- refactor helpers

### DIFF
--- a/packages/migrate/src/helpers/strings.ts
+++ b/packages/migrate/src/helpers/strings.ts
@@ -1,0 +1,8 @@
+/**
+ * Checks if there are changes in the code between to strings
+ * by comparing code without empty spaces between structures.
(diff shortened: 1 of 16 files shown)


packages/migrate/src/helpers/strings.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Checks if there are changes in the code between to strings
+ * by comparing code without empty spaces between structures.
+ */
+export function isSourceCodeChanged(originalText: string, updateText: string): boolean {
```
egorio, line 5: Refactoring. This function just compare strings... so moved it out if AST related helpers

PR #178, opened 2022-05-25
feat: Fix typescript error 2790 [PPTF-4317]

brendaHuang:
Error 2790 is the error we receive when we delete a property on a type or class that's required. The auto fix adopted in this PR is to make the property optional in the type definition or class definition. This PR handles when the type or class is local to the file where the error happens, or when the type or class is imported from a different file.

### DIFF
--- a/packages/migrate/src/helpers/transform-utils.ts
+++ b/packages/migrate/src/helpers/transform-utils.ts
@@ -1,0 +1,66 @@
+import ts from 'typescript';
+
+import type RehearsalService from '../rehearsal-service';
+
+export function getInterfaceByName(
+  sourceFile: ts.SourceFile,
+  typeName: string
+): ts.InterfaceDeclaration | undefined {
+  const matched = sourceFile.statements.find(
+    (s) => ts.isInterfaceDeclaration(s) && s.name.getFullText().trim() === typeName
+  );
+  return matched as ts.InterfaceDeclaration;
+}
+
+export function getImportByName(
+  sourceFile: ts.SourceFile,
+  typeName: string
+): ts.ImportDeclaration | undefined {
+  const matched = sourceFile.statements.find(
+    (s) =>
+      ts.isImportDeclaration(s) &&
+      (s.importClause?.namedBindings as ts.NamedImports).elements.find(
+        (e) => e.name.getFullText().trim() === typeName
+      )
+  );
+  return matched as ts.ImportDeclaration;
+}
+
+export function getClassByName(
+  sourceFile: ts.SourceFile,
+  className: string
+): ts.ClassDeclaration | undefined {
+  const matched = sourceFile.statements.find(
+    (s) => ts.isClassDeclaration(s) && s.name?.getFullText().trim() === className
+  );
+  return matched as ts.ClassDeclaration;
+}
+
+export function getSourceFileFromImport(
+  declaration: ts.ImportDeclaration,
+  importingFile: string,
+  service: RehearsalService
+): ts.SourceFile | undefined {
+  const module = declaration.moduleSpecifier.getFullText().trim().replace(/^'|'$/g, '');
+  const fileName = service.resolveModuleName(module, importingFile);
+  if (fileName) {
+    const sourceFile = service.getSourceFile(fileName);
+    return sourceFile;
+  }
+}
+
+export function getInterfaceMemberByName(
+  declaration: ts.InterfaceDeclaration,
+  typeMemberName: string
+): ts.TypeElement | undefined {
+  return declaration.members.find((member) => member.name?.getFullText().trim() === typeMemberName);
+}
+
+export function getClassMemberByName(
+  declaration: ts.ClassDeclaration,
+  classMemberName: string
+): ts.ClassElement | undefined {
+  return declaration.members.find(
+    (member) => member.name?.getFullText().trim() === classMemberName
+  );
+}
--- a/packages/migrate/src/helpers/typescript-ast.ts
+++ b/packages/migrate/src/helpers/typescript-ast.ts
@@ -106,3 +106,8 @@ export function isJsxTextNode(node: ts.Node): boolean {
 
   return visit(node);
 }
+
+export function insertIntoText(text: string, insertAt: number, strToInsert: string): string {
+  const newText = `${text.substring(0, insertAt)}${strToInsert}${text.substring(insertAt)}`;
+  return newText;
+}
--- a/packages/migrate/src/interfaces/fix-transform.ts
+++ b/packages/migrate/src/interfaces/fix-transform.ts
@@ -1,17 +1,19 @@
 import ts from 'typescript';
+import type RehearsalService from '../rehearsal-service';
 
 export interface FixedFile {
   fileName: string;
   text: string;
 }
+
 export default class FixTransform {
   /** Engineer friendly message describes steps needs to be done to fix the related issue */
   hint?: string;
 
   /** Function to fix the diagnostic issue */
-  fix?: (diagnostic: ts.DiagnosticWithLocation, service: ts.LanguageService) => FixedFile[];
+  fix?: (diagnostic: ts.DiagnosticWithLocation, service: RehearsalService) => FixedFile[];
 
-  run(diagnostic: ts.DiagnosticWithLocation, service: ts.LanguageService): FixedFile[] {
+  run(diagnostic: ts.DiagnosticWithLocation, service: RehearsalService): FixedFile[] {
     return this.fix ? this.fix(diagnostic, service) : [];
   }
 }
--- a/packages/migrate/src/transforms/2790-fix-transform.ts
+++ b/packages/migrate/src/transforms/2790-fix-transform.ts
@@ -1,0 +1,137 @@
+import ts from 'typescript';
+
+import FixTransform, { FixedFile } from '../interfaces/fix-transform';
+import { findNodeAtPosition, insertIntoText } from '../helpers/typescript-ast';
+import {
+  getClassByName,
+  getInterfaceByName,
+  getImportByName,
+  getSourceFileFromImport,
+  getClassMemberByName,
+  getInterfaceMemberByName,
+} from '../helpers/transform-utils';
+import type RehearsalService from '../rehearsal-service';
+
+const OPTIONAL_TOKEN = '?';
+
+export default class FixTransform2790 extends FixTransform {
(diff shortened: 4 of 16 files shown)


packages/migrate/src/interfaces/fix-transform.ts
```diff
@@ -1,17 +1,27 @@
 import ts from 'typescript';
+import RehearsalService from '../rehearsal-service';
```
lynchbomb, line 2: looks like this is being used only as a type. so instead just import the type eg. `import type RehearsalService from "../rehearsal-service";

packages/migrate/src/helpers/typescript-ast.ts
```diff
@@ -92,3 +92,25 @@ export function isSourceCodeChanged(originalText: string, updateText: string): b
...
+  const originalText = sourceFile.getFullText();
+  const newText = `${originalText.substring(0, insertAt)}${insertedStr}${originalText.substring(
+    insertAt
+  )}`;
+  const newSourceFile = ts.updateSourceFile(
+    sourceFile,
+    newText,
+    ts.createTextChangeRange({ start: 0, length: originalText.length }, newText.length)
+  );
+  return ts
```
egorio, line 110: Seems you don't need to work with SourceFile here. 
Can this function be just 
```
insertStringAtPosition(originalText, insertAt, insertedStr) {
    return `${originalText.substring(0, insertAt)}${insertedStr}${originalText.substring(insertAt)}`;
}
```
?

packages/migrate/src/transforms/2790-fix-transform.ts
```diff
@@ -0,0 +1,207 @@
...
+    (s) =>
+      ts.isImportDeclaration(s) &&
+      (s.importClause?.namedBindings as ts.NamedImports).elements.find(
+        (e) => e.name.getFullText().trim() === typeName
+      )
+  );
+  return matched as ts.ImportDeclaration;
+}
+
+function findLocalClass(
```
egorio, line 32: Probably findClassDeclaration ?

packages/migrate/src/transforms/2790-fix-transform.ts
```diff
@@ -0,0 +1,207 @@
...
+}
+
+function getSourceFileFromImport(
+  declaration: ts.ImportDeclaration,
+  basePath: string,
+  service: RehearsalService
+): ts.SourceFile {
+  const rawPath = declaration.moduleSpecifier.getFullText().trim();
+  const parts = rawPath
+    .replace(/^'(\.\/)?/, '') //remove quotation mark, and './' at the start
```
egorio, line 71: You don't need to remove '.' or '..' and split string. The [resolve](https://nodejs.org/api/path.html#pathresolvepaths) takes care of them.

This will work just fine:

```typescript
resolve(basePath, `relativeModulePath${.ts}`);
```

but seems this solution won't work perfectly [because](https://www.typescriptlang.org/docs/handbook/module-resolution.html) and there are different methods of resolving modules:

```
At this point, the compiler will ask “what’s the shape of moduleA?” While this sounds straightforward, moduleA could be defined in one of your own **.ts/.tsx** files, or in a **.d.ts** that your code depends on.
```
I think the right solution is using ts.resolveModuleName... 
You will probably don't need to use basePath in this case as well...
Let me create a PR with moduleResolution quickly... I played with it recently.

packages/migrate/src/transforms/2790-fix-transform.ts
```diff
@@ -0,0 +1,206 @@
...
+
+import path from 'path';
+
+import FixTransform, { FixedFile } from '../interfaces/fix-transform';
+import { findNodeAtPosition, insertStringIntoSourceFile } from '../helpers/typescript-ast';
+import RehearsalService from 'src/rehearsal-service';
+
+const QUESTION_MARK = '?';
+
+function findLocalInterface(
```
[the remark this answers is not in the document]
asakusuma: I think it could make sense to pull out into a separate package. I would love to use these utils for linters and codemods. I do have a hard time believing these utils don't already exist somewhere. Like VScode essentially does this same process for "Go to Definition" and I'm sure linters exist that do the same thing.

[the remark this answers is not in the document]
asakusuma: @brendaHuang perhaps look for a linter or codemod that already exists that will go and find the source of a type in another file?

packages/migrate/src/interfaces/fix-transform.ts
```diff
@@ -1,17 +1,27 @@
...
+    diagnostic: ts.DiagnosticWithLocation,
+    bathPath: string,
+    service: RehearsalService
+  ) => FixedFile[];
 
-  run(diagnostic: ts.DiagnosticWithLocation, service: ts.LanguageService): FixedFile[] {
-    return this.fix ? this.fix(diagnostic, service) : [];
+  run(
+    diagnostic: ts.DiagnosticWithLocation,
+    basePath: string,
```
egorio, line 22: I feel the `service` is a good place to hold `basePath`... can event be implemented as a getter to grab the "calculated" base path from compilerOptions.

packages/migrate/src/interfaces/fix-transform.ts
```diff
@@ -1,17 +1,27 @@
...
+
 export default class FixTransform {
   /** Engineer friendly message describes steps needs to be done to fix the related issue */
   hint?: string;
 
   /** Function to fix the diagnostic issue */
-  fix?: (diagnostic: ts.DiagnosticWithLocation, service: ts.LanguageService) => FixedFile[];
+  fix?: (
+    diagnostic: ts.DiagnosticWithLocation,
+    bathPath: string,
```
egorio, line 16: Typo :)

packages/migrate/test/fixtures/migrate/2790-import-2.ts.output
```diff
@@ -0,0 +1,11 @@
+export class Employee {
+    name: string;
+    badgeNumber?: string;
+    constructor({ name, badgeNumber }: {
```
egorio, line 4: Output has broken formatting - all empty lines been removed. 
This can be issue of using ts printer... or the way plugins run currently.

packages/migrate/test/fixtures/migrate/2790-import-1.ts.output
```diff
@@ -0,0 +1,9 @@
+export interface Animal {
+    species: string;
+    weight?: number;
+}
+export interface Car {
```
egorio, line 5: The same as below - format is broken.

packages/migrate/src/transforms/2790-fix-transform.ts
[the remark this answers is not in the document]
egorio: @asakusuma Good point.
@brendaHuang Let's keep helpers in migrate for now, will move the out in separate PR (PPTF-4349).

packages/migrate/src/transforms/2790-fix-transform.ts
brendaHuang: Changed to service.getResolveModuleName()

packages/migrate/test/fixtures/migrate/2790-import-2.ts.output
brendaHuang: Good catch. Removed the printer and the formatting is fixed.

packages/migrate/src/interfaces/fix-transform.ts
brendaHuang: changed

packages/migrate/src/transforms/2790-fix-transform.ts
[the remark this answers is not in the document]
brendaHuang: sounds good

packages/migrate/src/helpers/typescript-ast.ts
brendaHuang: Removed printer

packages/migrate/src/transforms/2790-fix-transform.ts
brendaHuang: changed

packages/migrate/src/transforms/2790-fix-transform.ts
```diff
@@ -0,0 +1,207 @@
...
+  sourceFile: ts.SourceFile,
+  typeName: string
+): ts.InterfaceDeclaration | undefined {
+  const matched = sourceFile.statements.find(
+    (s) => ts.isInterfaceDeclaration(s) && s.name.getFullText().trim() === typeName
+  );
+  return matched as ts.InterfaceDeclaration;
+}
+
+function findImport(sourceFile: ts.SourceFile, typeName: string): ts.ImportDeclaration | undefined {
```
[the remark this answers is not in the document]
brendaHuang: changed function name

packages/migrate/src/interfaces/fix-transform.ts
brendaHuang: Removed basePath.

packages/migrate/src/interfaces/fix-transform.ts
brendaHuang: :-)

packages/migrate/test/fixtures/migrate/2790-import-1.ts.output
brendaHuang: fixed

packages/migrate/src/helpers/transform-utils.ts
```diff
@@ -0,0 +1,66 @@
...
+  sourceFile: ts.SourceFile,
+  className: string
+): ts.ClassDeclaration | undefined {
+  const matched = sourceFile.statements.find(
+    (s) => ts.isClassDeclaration(s) && s.name?.getFullText().trim() === className
+  );
+  return matched as ts.ClassDeclaration;
+}
+
+export function getSourceFileFromImport(
```
egorio, line 39: Just to note: this function, probably, should be a part of @rehearsla/utils package because it depends on RehearsalService

brendaHuang: If we separately pass in RehearsalService's language service host and its language service, this function can be decoupled from RehearsalService. But now both are protected member variables of the RehearsalService, and cannot be accessed in this manner. So tradeoffs for the current design of the RehearsalService.

egorio (approved):

PR #180, opened 2022-06-01
Fix: undefined members in import clause

brendaHuang:


.vscode/launch.json
```diff
@@ -78,5 +78,25 @@
...
+      "runtimeArgs": ["--nolazy", "-r", "ts-node/register"],
+      "args": [
+        "--no-timeout",
+        "${workspaceFolder}/packages/migrate/test/migrate.test.ts"
+      ],
+      "sourceMaps": true,
+      "cwd": "${workspaceFolder}/packages/migrate",
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "protocol": "inspector",
```
lynchbomb, line 95: I think this property is deprecated?

lynchbomb (changes requested):
Minor fixes

.vscode/launch.json
brendaHuang: Listed here: https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_javascript-debug-terminal
But listed with attributes for 'attach'. So removing it.

.vscode/launch.json
```diff
@@ -78,5 +78,25 @@
...
     },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Only Migrate",
+      "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
+      "runtimeArgs": ["--nolazy", "-r", "ts-node/register"],
+      "args": [
+        "--no-timeout",
+        "${workspaceFolder}/packages/migrate/test/migrate.test.ts"
```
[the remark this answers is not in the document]
brendaHuang: changed

packages/migrate/src/helpers/transform-utils.ts
```diff
@@ -12,16 +12,31 @@ export function getInterfaceByName(
   return matched as ts.InterfaceDeclaration;
 }
 
+export function hasNameInImportClause(
+  importClause: ts.ImportClause | undefined,
+  importName: string
+): boolean {
+  if (importClause?.name?.getFullText().trim() === importName) {
+    return true;
```
egorio, line 20: Can you add test case for this, please. 
And should we cover `import * from "something"` as well?

brendaHuang: That will be the complete solution I mentioned early, the test cases and other things. Accessing the properties for such use cases are quirky compared to when the imports are in curly braces.

packages/migrate/src/helpers/transform-utils.ts
```diff
@@ -12,16 +12,31 @@ export function getInterfaceByName(
...
 ): ts.ImportDeclaration | undefined {
   const matched = sourceFile.statements.find(
-    (s) =>
-      ts.isImportDeclaration(s) &&
-      (s.importClause?.namedBindings as ts.NamedImports).elements.find(
-        (e) => e.name.getFullText().trim() === typeName
-      )
+    (s) => ts.isImportDeclaration(s) && hasNameInImportClause(s.importClause, typeName)
   );
   return matched as ts.ImportDeclaration;
```
egorio, line 41: Will this fail if `matched` is undefined?

brendaHuang: No. The return type of this function is specified as ts.ImportDeclaration | undefined. The function that calls this function handles both.

PR #205, opened 2022-06-14
feat: fix TS2345

brendaHuang:
This PR is to address the following error from nuage-core. 

`src/components/FeedbackDialog/index.tsx:234:22 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'SetStateAction<string>'.

234         setJiraError(error);
                         ~~~~~`

### DIFF
--- a/packages/migrate/src/plugins/diagnostics-autofix.plugin.ts
+++ b/packages/migrate/src/plugins/diagnostics-autofix.plugin.ts
@@ -8,6 +8,7 @@ import {
   FixTransform2571,
   FixTransform2790,
   FixTransform6133,
+  FixTransform2345,
 } from '../transforms';
 
 import Plugin, { type PluginParams, type PluginResult } from '../interfaces/plugin';
@@ -33,7 +34,7 @@ export default class DiagnosticAutofixPlugin extends Plugin {
 
       const fixedFiles = fix.run(diagnostic, this.service);
 
-      const hint = this.prepareHint(diagnostic.messageText, fix?.hint);
+      const hint = this.prepareHint(diagnostic.messageText, diagnostic, fix?.getHint);
 
       const fixed = fixedFiles.length > 0;
 
@@ -65,24 +66,30 @@ export default class DiagnosticAutofixPlugin extends Plugin {
   /**
    * Prepares a hint message for engineer based on original diagnostic message and `this.hint`
    */
-  prepareHint(message: string | ts.DiagnosticMessageChain, hint?: string): string {
+  prepareHint(
+    message: string | ts.DiagnosticMessageChain,
+    diagnostic: ts.DiagnosticWithLocation,
+    getHint?: (
+      replacements: { [key: string]: string },
+      tsDiagnostic: ts.DiagnosticWithLocation
+    ) => string | undefined
+  ): string {
     message = ts.flattenDiagnosticMessageText(message, '. ');
+    let hint: string | undefined;
+    // Prepare a replacement dictionary
+    // e.g. {'{0}': 'p', '{1}': 'any'} for message "Parameter 'p' implicitly has an 'any' type."
+    const replacements: { [key: string]: string } =
+      message
+        // return ["'p'", "'any'"]
+        .match(/'[^']+'/gm)
+        // converts ["'p'", "'any'"] to {'{0}': 'p', '{1}': 'any'}
+        ?.reduce((a, v, i) => ({ ...a, [`{${i}}`]: v.replace(/^\W+|\W+$/g, '') }), {}) || {};
 
-    if (hint) {
-      // Prepare a replacement dictionary
-      // e.g. {'{0}': 'p', '{1}': 'any'} for message "Parameter 'p' implicitly has an 'any' type."
-      const replacements: { [key: string]: string } =
-        message
-          // return ["'p'", "'any'"]
-          .match(/'[^']+'/gm)
-          // converts ["'p'", "'any'"] to {'{0}': 'p', '{1}': 'any'}
-          ?.reduce((a, v, i) => ({ ...a, [`{${i}}`]: v.replace(/^\W+|\W+$/g, '') }), {}) || {};
-
-      // Replaces {0}, {1}, ... placeholders with corresponding values from the original message
-      message = hint.replace(/{\d+}/gm, (key) => replacements[key] || key);
+    if (getHint) {
+      hint = getHint(replacements, diagnostic);
     }
 
-    return message;
+    return hint ? hint : message;
   }
 
   /**
@@ -115,6 +122,7 @@ export function getFixForDiagnostic(diagnostic: ts.Diagnostic): FixTransform {
     2571: FixTransform2571,
     2790: FixTransform2790,
     6133: FixTransform6133,
+    2345: FixTransform2345,
   };
 
   return diagnostic.code in availableFixes
--- a/packages/migrate/src/transforms/2345-fix-transform.ts
+++ b/packages/migrate/src/transforms/2345-fix-transform.ts
@@ -1,0 +1,25 @@
+// import ts from 'typescript';
+
+import FixTransform, { FixedFile } from '../interfaces/fix-transform';
+// import { findNodeAtPosition } from '../helpers/typescript-ast';
+
+export default class FixTransform2345 extends FixTransform {
+  getHint = (replacements: { [key: string]: string }) => {
+    if (Object.values(replacements).length < 2) {
+      return undefined;
+    }
+    let hint = '';
+
+    const { '{0}': actualType, '{1}': expectedType } = replacements;
+    if (actualType === 'unknown' || actualType === 'any') {
+      hint = `Argument of type '${actualType}' is not assignable to parameter of type '${expectedType}'. Consider specifying type of argument to be '${expectedType}', using type assertion: '(argument as ${expectedType})', or using type guard: 'if (argument instanceof ${expectedType}) { ... }'.`;
+    } else {
+      hint = `Argument of type '${actualType}' is not assignable to parameter of type '${expectedType}'. Consider verifying both types, using type assertion: '(argument as ${expectedType})', or using type guard: 'if (argument instanceof ${expectedType}) { ... }'.`;
+    }
+    return hint;
+  };
+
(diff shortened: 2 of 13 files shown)


packages/migrate/src/transforms/2345-fix-transform.ts
```diff
@@ -0,0 +1,28 @@
+import ts from 'typescript';
+
+import FixTransform, { FixedFile } from '../interfaces/fix-transform';
```
lynchbomb, line 3: import FixTransform, { type FixedFile } from '../interfaces/fix-transform';

packages/migrate/src/transforms/2345-fix-transform.ts
```diff
@@ -0,0 +1,28 @@
...
+  hint = `Argument of type '{0}' is not assignable to parameter of type '{1}'.`;
+
+  fix = (diagnostic: ts.DiagnosticWithLocation): FixedFile[] => {
+    const errorNode = findNodeAtPosition(diagnostic.file, diagnostic.start, diagnostic.length);
+    const variableName = errorNode?.getText().trim();
+    const message = ts.flattenDiagnosticMessageText(diagnostic.messageText, '. ');
+    const types = message.match(/'([^']+)'/gm)?.map((matched) => matched.replace(/^'|'$/g, ''));
+    if (Array.isArray(types) && variableName) {
+      const [actualType, expectedType] = types;
+      if (actualType === 'unknown' || actualType === 'any') {
```
lynchbomb, line 16: If this use case goes beyond the current implementation of [prepareHint](https://github.com/rehearsal-js/rehearsal-js/blob/948eb8a8dea1aae76f26d82e4cf39f919ffd68c4/packages/migrate/src/plugins/diagnostics-autofix.plugin.ts#L68) we should consider expanding the method to handle this use case. 

Or just do a positional replace on the delta of the help message. Looks like the delta is only `specifying type of argument to be ${expectedType}`. eg

hint = `Argument of type '{0}' is not assignable to parameter of type '{1}'. '{2}', using type assertion: '(${variableName} as ${expectedType})', or using type guard: 'if (${variableName} instanceof ${expectedType}) { ... }'.`

lynchbomb (changes requested):
Looks good. Left a few minor comments on simplifying.

packages/migrate/src/transforms/2345-fix-transform.ts
brendaHuang: FixedFile itself is an interface. Do I need to say 'type FixedFile' while importing it?

brendaHuang:
@lynchbomb @egorio  Changes in this commit: 

1.  Creates a function `getHint()` to get the customized hint for each autofix, since we need logic and access to diagnostic nodes to create hints. The `prepareHint` function in the plugin file takes the customized getHint() function instead of a hint string.  
2. Separate the `fix()` function from the `getHint()` function so that the former focuses on AST modification and the latter focuses on creating customized hints. Previously the two functionalities were mixed in certain cases, for example we used `fix()` to just create hints because we need access to the diagnostic node.

packages/migrate/src/transforms/2345-fix-transform.ts
egorio: Hint intentionally left simple to keep all conditions in fix function. For some condition fixes can be applied in a first place and for other - only hints. Later hints can be replaced with auto fixes, but no conditions will be changed

egorio:
Brenda, could you please check https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migrate/src/transforms/6133-fix-transform.ts. This case was already covered and you don’t need getHint function.

packages/migrate/src/transforms/2345-fix-transform.ts
brendaHuang: @lynchbomb @lynchbomb My first version of TS2345 is to create the hint in the `fix()` method. I can see the merits of both approaches: merging the hint and the autofix in one `fix()` function or separate functions for each. 

The first approach may be for performance in certain cases when both autofix and hints share access to the diagnostic node. 

The second approach may be better for readability and maintenance. In addition, the fix() method returns the files that have been changed and the plugin file uses that value to report whether any autofix has been done. If the fix() is also responsible for producing hints, then it's a function that does two things at one time. And the need for shared access to the diagnostic node doesn't happen each time when creating hints and modifying AST. 

Each has tradeoffs. I am OK with either way.

packages/migrate/src/transforms/6133-fix-transform.ts
```diff
@@ -2,24 +2,35 @@ import ts from 'typescript';
 
 import FixTransform, { FixedFile } from '../interfaces/fix-transform';
 import { isSourceCodeChanged } from '../helpers/strings';
-import { transformDiagnosedNode } from '../helpers/typescript-ast';
+import { transformDiagnosedNode, findNodeAtPosition } from '../helpers/typescript-ast';
 
 export default class FixTransform6133 extends FixTransform {
-  hint = `The declaration '{0}' is never read or used. Remove the declaration or use it.`;
+  getHint = (replacements: { [key: string]: string }, diagnostic: ts.DiagnosticWithLocation) => {
```
egorio, line 8: This makes fix inefficient because we have to traverse the AST tree twice - in getHint (findNodeAtPosition) and in fix (transformDiagnosedNode). Plus makes code much complicated (lines 9 - 15 not needed in the original solution)

packages/migrate/src/transforms/2790-fix-transform.ts
```diff
@@ -17,7 +17,14 @@ import type RehearsalService from '../rehearsal-service';
 const OPTIONAL_TOKEN = '?';
 
 export default class FixTransform2790 extends FixTransform {
-  hint = `The operand of a 'delete' operator must be optional.`;
+  getHint = (replacements: { [key: string]: string }) => {
```
egorio, line 20: You can clearly see the second approach just made code complicated - we have to "compute" hint in every getHint, instead of keep this logic outside (originally).

packages/migrate/src/transforms/2571-fix-transform.ts
```diff
@@ -4,7 +4,15 @@ import { isVariableOfCatchClause, transformDiagnosedNode } from '../helpers/type
...
 
 export default class FixTransform2571 extends FixTransform {
-  hint = `Object is of type '{0}'. Specify a type of variable, use type assertion: \`(variable as DesiredType)\` or type guard: \`if (variable instanceof DesiredType) { ... }\``;
+  getHint = (replacements: { [key: string]: string }) => {
+    if (Object.values(replacements).length < 1) {
+      return undefined;
+    }
+
+    const { '{0}': actualType } = replacements;
+    const hint = `Object is of type '${actualType}'. Specify a type for variable, use type assertion: \`(variable as DesiredType)\` or type guard: \`if (variable instanceof DesiredType) { ... }\``;
```
egorio, line 13: Originally, kind of a code snippet, was provided, because real variable name was used.
The message (variable as DesiredType) is less useful than (rean_variable_name as DesiredType).

packages/migrate/src/plugins/diagnostics-autofix.plugin.ts
```diff
@@ -65,24 +66,30 @@ export default class DiagnosticAutofixPlugin extends Plugin {
...
+    // e.g. {'{0}': 'p', '{1}': 'any'} for message "Parameter 'p' implicitly has an 'any' type."
+    const replacements: { [key: string]: string } =
+      message
+        // return ["'p'", "'any'"]
+        .match(/'[^']+'/gm)
+        // converts ["'p'", "'any'"] to {'{0}': 'p', '{1}': 'any'}
+        ?.reduce((a, v, i) => ({ ...a, [`{${i}}`]: v.replace(/^\W+|\W+$/g, '') }), {}) || {};
+
+    if (getHint) {
+      hint = getHint(replacements, diagnostic);
```
egorio, line 89: The benefit of original code is to have simple transforms, since all replacement logic is happening outside.
With getHint, engineer have "repeat" this logic in every function :(

packages/migrate/src/transforms/2345-fix-transform.ts
egorio: > The first approach may be for performance in certain cases when both autofix and hints share access to the diagnostic node.

There is no difference in performance in case you don't need a node for message (you don't required to traverse), but there is, actually, decrease performance [in case](https://github.com/rehearsal-js/rehearsal-js/pull/205/files#diff-224757bec7f6966cf22692d794cec47df113b520f11fc1b6dddb9df97712ecc6R8-R35) of having both fix (for some cases) and hint (for other) because you need to traverse tree twice - first in fix(), second in getHint().

Since, the target is to provide better hints (even with some code snippets), the node [is usually needed](https://github.com/rehearsal-js/rehearsal-js/pull/205/files#diff-c0bb29af946a24a72228039bfe46599f98193e72d82154391159b2997b10426aL18). So, we need to traverse anyway. 

> The second approach may be better for readability ....

Yeah, this is known issue, but there is a plan to address it a different way.
Renaming function fix() to something else to highlight this is not just a fix, but to be an entry point entry point / controller for autofix class, which will be called fixOneCase(), fixOtherCase(), ... or provideHint() depends on conditions.

> and maintenance.

For this purpose fix should only have condition and either call appropriate fix function or change hint (or, probably,  even both in future.

Like:
```
const text = transformDiagnosedNode(diagnostic, (node: ts.Node) => {
      if (ts.isImportDeclaration(node) || ts.isImportSpecifier(node)) {
        this.removeImportDeclaration()
      } else if (ts.isIdentifier(node) && ts.isFunctionDeclaration(node.parent)) {
        this.RepoveFunctionDeclaration()
      } else if (ts.isIdentifier(node) && ts.isParameter(node.parent)) {
        this.hint = `The function '{0}' is never called. Remove the function or use it.`;
        // Or even this.setHint(`The function '{0}' is never called. Remove the function or use it.`;)
      }

      return node;
    });
```

There is no benefits of using `findNodeAtPosition` over `transformDiagnosedNode` because both of them do the same, except transformDiagnosedNode fires callback while findNodeAtPosition you have to do perform additional checks for the node manually. 

> Each has tradeoffs. I am OK with either way.

PR #210, opened 2022-06-22
feat: add details hint to TS2345

brendaHuang:
This PR is to address the following error from nuage-core.

`src/components/FeedbackDialog/index.tsx:234:22 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'SetStateAction'.

234 setJiraError(error);
~~~~~`

### DIFF
--- a/packages/migrate/src/helpers/transform-utils.ts
+++ b/packages/migrate/src/helpers/transform-utils.ts
@@ -71,6 +71,21 @@ export function getTypeNameFromType(
   return typeName;
 }
 
+export function getTypeNameFromVariable(node: ts.Node, program: ts.Program): string | undefined {
+  const checker = program.getTypeChecker();
+  const type = checker.getTypeAtLocation(node);
+  if (!type) {
+    return undefined;
+  }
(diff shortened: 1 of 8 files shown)


packages/migrate/src/helpers/transform-utils.ts
```diff
@@ -71,6 +71,21 @@ export function getTypeNameFromType(
...
+  const checker = program.getTypeChecker();
+  const type = checker.getTypeAtLocation(node);
+  if (!type) {
+    return undefined;
+  }
+  if (type.isNumberLiteral()) {
+    return 'number';
+  }
+  if (type.isStringLiteral()) {
+    return 'string';
```
egorio, line 84: Just wondering if it make sense to have enum for types (for future fixes)?

lynchbomb (approved):

packages/migrate/src/helpers/transform-utils.ts
brendaHuang: sounds good

PR #269, opened 2022-07-27
chore: Refactor Reporter: better naming and no default exports

egorio:
Remove default export
Rename pullRequestMd to mdFormatter, because it just an example of md formatter (not specifically for PRs)


packages/reporter/src/index.ts
```diff
@@ -1,6 +1,7 @@
-export { default } from './reporter';
```
egorio, line 1: Keep only one default export per package

lynchbomb (approved):

PR #284, opened 2022-08-04
Create migrate module; Refactor out shared code between migrate and upgrade modules

wondersloth:
# Summary of Features
- This builds out the migrate module. 
  - Will convert a `.js` to `.ts` file.
  - Will add types that it can infer to values.
  - Will add types from `.ts` files in the MigrateInput object.
- Created a shared package for `plugins`
  - Refactored FixTransform interfaces.
- Moved `helpers` from upgrade to `utils`

# Testing
- Setup `@reharsal/migrate` to use vitest.
- Wrote acceptance test for validating that types get added to a file that is converted from js to ts.


packages/shared/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@rehearsal/shared",
```
egorio, line 2: Could you rename package to `@rehearsal/plugins` + move files under `helpers/` to `@rehearsal/utils`?

packages/migrate/package.json
```diff
@@ -21,22 +21,25 @@
...
   "devDependencies": {
-    "@types/chai": "^4.3.0",
     "@types/eslint": "^8.4.1",
-    "@types/mocha": "^9.1.0",
     "@types/node": "^18.0.3",
     "@types/react": "^18.0.9",
-    "chai": "^4.3.6",
-    "mocha": "^10.0.0",
-    "nyc": "^15.1.0"
+    "vitest": "^0.16.0"
```
egorio, line 42: Why vitest? I thought to migrate all packages to something... probably vitest is the good candidate?

wondersloth: Mark suggested it as for an alternative testing tool. I used it in the `migration-graph` library.

It's pretty great.
- Does easy compilation of all `.ts` files.
- Allows for all test files to be in `.ts`
- Does hot-reload of test files with the watcher in dev mode.
- The API is compatible with jest tests as well. 

In general I think this will be a great library to migrate to for internal `play/glimmer` applications since a large portion of them use Jest.

egorio: Cool. I'll migrate other packages slowly then :)

packages/shared/package.json
wondersloth: Thanks for suggestion. Having a plugins package we can create a suite of tools around testing too.

packages/migrate/package.json
lynchbomb: Eventually we should migrate all packages to vitest.

egorio (approved):

packages/migrate/src/transforms/7006-fix-transform.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export class FixTransform7006 extends FixTransform {
+  hint = `TBD Parameter '{0}' implicitly has an '{1}' type.`; // TODO: this is not final
+
+  fix = (diagnostic: ts.DiagnosticWithLocation, service: RehearsalService): FixResult => {
+    const node = findNodeAtPosition(diagnostic.file, diagnostic.start, diagnostic.length);
+
+    if (!node /* || !ts.isIdentifier(errorNode) */) {
+      // TODO: Does this need to be uncommented?
+      return { fixedFiles: new Array<FixedFile>(), commentedFiles: new Array<FixedFile>() };
```
brendaHuang, line 15: I am working on changing the type of FixResult. Using the new type, this line would not mark the result as commented. It will just return a blank FixResult template showing that the file has been inspected but nothing has been done.

packages/migrate/src/transforms/7006-fix-transform.ts
```diff
@@ -0,0 +1,93 @@
...
+    const fixes = languageService.getCodeFixesAtPosition(
+      sourceFile.fileName,
+      diagnostic.start,
+      diagnostic.start + diagnostic.length,
+      [7006],
+      formatOptions,
+      {}
+    );
+
+    const stubLocation = { line: 0, character: 0 }; // TODO: actually get the correct coordinates for this.
```
brendaHuang, line 35: By knowing the value of `diagnostic.start` and using `ts.getLineAndCharacterOfPosition()`, we can easily get the location.

brendaHuang (approved):

packages/migrate/src/transforms/7006-fix-transform.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    const languageService = service.getLanguageService();
+    const formatOptions = ts.getDefaultFormatCodeSettings();
+    const sourceFile = diagnostic.file;
+
+    const fileName = sourceFile.fileName;
+
+    const results: Array<FixedFile> = new Array<FixedFile>();
+
+    const fixes = languageService.getCodeFixesAtPosition(
```
brendaHuang, line 26: If service.getCodeFixesAtPosition() works for a significant amount of scenarios, the fix may later needs to be extracted and used for other TS errors.

PR #293, opened 2022-08-11
chore: Migrate tests to vitest, except cli (#287)

egorio:
Migrate tests to vitest, except cli because of https://github.com/rehearsal-js/rehearsal-js/pull/291

Renamed test files of utils because of naming convention. We can review our convention later.


haochuan (approved):
Looks good!

PR #311, opened 2022-08-25
chore: refactor sarif formatter

chadhietala:
This refactors the sarif formatter to not rely on module scope state. In general we want to avoid manually GCing objects by constructing code in a way where once the objects out out of scope the GC will take care of the objects for us.


egorio:
Thanks @chadhietala. I didn't think about GC before. Seems it makes sense to refactor [CodeFixes](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/upgrade/src/transforms/4082/index.ts#L65-L174) the same way :)

egorio (approved):

PR #313, opened 2022-08-29
fix: Fix the duplicate comments issue

### LINKED ISSUE #207: Duplicated Comments
When applying transforms previously included rehearsal comment lines are appended rather than replaced. These lines should be updated if a new hint etc. is available for a given diagnostic.

egorio:
Add the `Re-Rehearse` plugin which is responsible to clean up previously added `@rehearsal` comments on re-run.

Upgrade `Diagnostics-Autofix` plugin to make sure it skips all previously processed error based on `@rehearal` comment without dependency on `@ts-ignore` comment which doesn't always suspend errors as expected. 

Fixes #207.


packages/plugins/src/plugins/re-rehearsal.plugin.ts
```diff
@@ -0,0 +1,48 @@
...
+} from '@rehearsal/service';
+
+/**
+ * Removes all comments with `@rehearsal` tag inside
+ */
+export class ReRehearsalPlugin extends Plugin {
+  async run(params: PluginParams<undefined>): PluginResult {
+    const { fileName } = params;
+
+    const tag = '@rehearsal';
```
egorio, line 16: This will be moved to plugin' options later - https://github.com/rehearsal-js/rehearsal-js/issues/314

packages/plugins/tsconfig.json
```diff
@@ -5,4 +5,4 @@
     "outDir": "dist"
   },
   "include": ["src"]
-}
```
lynchbomb, line 8: I dont see a diff here likely just reset this file

lynchbomb (approved):

PR #327, opened 2022-09-01
refactor: extract transforms into new package

lynchbomb:
This is a large refactor PR which at its core creates a new package for all `transforms` and removes some redundancy across packages and cleans up import/exports.


packages/transforms/src/fix-transform.ts
```diff
@@ -0,0 +1,57 @@
...
+  updatedText: string;
+  code?: string;
+  codeFixAction?: CodeFixAction;
+  location: {
+    line: number;
+    character: number;
+  };
+}
+
+export interface FixTransformI {
```
egorio, line 18: This interface seems redundant

packages/transforms/src/fix-transform.ts
```diff
@@ -0,0 +1,57 @@
...
+  /** Engineer friendly message describes steps needs to be done to fix the related issue */
+  hint?: string;
+
+  /** Function to fix the diagnostic issue */
+  fix?: (diagnostic: ts.DiagnosticWithLocation, service: RehearsalService) => FixedFile[];
+
+  run(diagnostic: ts.DiagnosticWithLocation, service: RehearsalService): FixedFile[];
+}
+
+export class FixTransform implements FixTransformI {
```
egorio, line 28: We're planning to rename this to `Codefix` (`Codefix1234`), but probably it's better to have separate PR for this :)

packages/upgrade/package.json
```diff
@@ -34,6 +34,7 @@
     "@rehearsal/reporter": "^0.0.30",
     "@rehearsal/service": "^0.0.30",
     "@rehearsal/utils": "^0.0.30",
+    "@rehearsal/transforms": "^0.0.30",
```
egorio, line 37: Could you place rename package to `@rehearsal/codefixes`, the transforms are only parts of a codefixes, the same as a hint :)

PR #330, opened 2022-09-01
test: expand migrate test to check for report

lynchbomb:


packages/migrate/tests/index.test.ts
```diff
@@ -21,21 +21,21 @@ describe('migrate', () => {
   function clean(dir: string): void {
     fs.rmSync(dir, { recursive: true, force: true });
   }
-
+  // copy the fixture files to a tmp directory
   function prepareInputFiles(files: string[] = ['index.js']): string[] {
```
egorio, line 25: We have helper in utils https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/plugins/test/re-rehearse.plugin.test.ts#L18

egorio: https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/utils/src/tests.utils.ts

haochuan (approved):
Cool! I'm going to double check the report in `cli`

Issue #314, opened 2022-08-29
Consider to make plugins configurable


egorio:

egorio:
Don't need for now, can consider create similar task in future

PR #349, opened 2022-09-02
refactor: Remove DiscoverTypesPlugin, PluginParams and add DiagnosticWithContext, add PluginOptions

egorio:
Remove DiscoverTypesPlugin, since it works the same as DiagnosticFixPlugin (we can consider to use different lists of codefixes if needed)

Remove PluginParams, because we have only one param when running plugins, but we will need [options](https://github.com/rehearsal-js/rehearsal-js/issues/314) when construct plugin object 

Created a DiagnosticWithContext interface to pass the diagnostic and program context downstream for easy access


lynchbomb:
This looks good overall. Once the failing tests are squared away ship-it.

lynchbomb (approved):

egorio:
> Once the failing tests are squared away ship-it.

We need to add plugin options to pas tests :)) 
Will do it a bit later.

PR #352, opened 2022-09-06
feat: new flag -r to generate report

haochuan:
- Just figured out I need to call `Reporter.save` to have the actual report, adding a new flag `-r` for generating report
- Renamed the previous option `root` to `basePath`


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "0.0.30",
```
egorio, line 2: You don't need to update version manually. Lerna does this on `lerna publish`

egorio:
The reporter technically can produce several different reports (json, serif) together. 
It's probably can be better to have something like `-r=json`, `-r=json,serif`... but can be upgraded later.
I would also recommend to add `-o` to specify reports output directory (planning to add it upgrade command)

haochuan:
@egorio Thanks for the detailed info and yeah all those makes sense. I'm not aware of the different report format before, will update this PR for that.

lerna.json
haochuan: hmmm wired I haven't touched this file manually.

packages/cli/src/commands/migrate.ts
```diff
@@ -44,11 +45,18 @@ type ParsedModuleResult = {
...
+type FormatterMap = {
+  [format: string]: FormatterFunction;
+};
+
 migrateCommand
   .description('Migrate Javascript project to Typescript')
   .requiredOption('-b, --basePath <project base path>', 'Base dir path of your project.')
   .requiredOption('-e, --entrypoint <entrypoint>', 'entrypoint js file for your project')
-  .option('-r, --report', 'create report for migration result')
+  .option('-r, --report <reportFormat>', 'Report formats separated by comma, e.g. -r json,sarif,md')
```
egorio, line 58: You can use [variadic option](https://www.npmjs.com/package/commander#variadic-option) or see [commaSeparatedList](https://www.npmjs.com/package/commander#custom-option-processing)

lerna.json
```diff
@@ -1,5 +1,5 @@
...
+  "version": "0.0.31",
```
lynchbomb, line 2: revert

packages/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rehearsal/cli",
-  "version": "0.0.30",
+  "version": "0.0.31",
```
lynchbomb, line 3: revert

packages/cli/package.json
```diff
@@ -35,9 +35,9 @@
   },
   "dependencies": {
     "@microsoft/api-documenter": "^7.13.68",
-    "@rehearsal/migrate": "^0.0.30",
-    "@rehearsal/reporter": "^0.0.30",
-    "@rehearsal/upgrade": "^0.0.30",
+    "@rehearsal/migrate": "^0.0.31",
```
lynchbomb, line 38: revert all 3 of these

packages/cli/src/commands/migrate.ts
lynchbomb: agreed. I wonder if we can do the same with the flag directions eg. `Report formats separated by comma, e.g. -r <reportTypes>`

packages/codefixes/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rehearsal/codefixes",
-  "version": "0.0.30",
+  "version": "0.0.31",
```
lynchbomb, line 3: revert

packages/codefixes/package.json
```diff
@@ -31,8 +31,8 @@
     "test:watch": "vitest --coverage --watch"
   },
   "dependencies": {
-    "@rehearsal/service": "^0.0.30",
-    "@rehearsal/utils": "^0.0.30"
+    "@rehearsal/service": "^0.0.31",
```
lynchbomb, line 34: revert both of these

packages/migrate/package.json
```diff
@@ -30,10 +30,10 @@
     "test:watch": "vitest --coverage --watch"
   },
   "dependencies": {
-    "@rehearsal/plugins": "^0.0.30",
-    "@rehearsal/reporter": "^0.0.30",
-    "@rehearsal/service": "^0.0.30",
-    "@rehearsal/utils": "^0.0.30",
+    "@rehearsal/plugins": "^0.0.31",
```
lynchbomb, line 33: revert all 4 of these

packages/migrate/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rehearsal/migrate",
-  "version": "0.0.30",
+  "version": "0.0.31",
```
lynchbomb, line 3: revert

packages/migration-graph-ember/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rehearsal/migration-graph-ember",
-  "version": "0.0.30",
+  "version": "0.0.31",
```
lynchbomb, line 3: revert

packages/plugins/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rehearsal/plugins",
-  "version": "0.0.30",
+  "version": "0.0.31",
```
lynchbomb, line 3: revert

packages/migration-graph/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rehearsal/migration-graph",
-  "version": "0.0.30",
+  "version": "0.0.31",
```
lynchbomb, line 3: revert

packages/reporter/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rehearsal/reporter",
-  "version": "0.0.30",
+  "version": "0.0.31",
```
lynchbomb, line 3: revert

packages/service/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rehearsal/service",
-  "version": "0.0.30",
+  "version": "0.0.31",
```
lynchbomb, line 3: revert

lynchbomb (changes requested):

packages/cli/src/commands/migrate.ts
haochuan: I went with [Custom option processing](https://www.npmjs.com/package/commander#custom-option-processing) to handle the comma-separated args, but open to change to `variadic` if you guys prefer it.

packages/cli/src/commands/migrate.ts
```diff
@@ -45,12 +46,23 @@ type ParsedModuleResult = {
...
+
+type FormatterMap = {
+  [format: string]: FormatterFunction;
+};
+
 migrateCommand
   .name('migrate')
   .description('Migrate Javascript project to Typescript')
-  .requiredOption('-r, --root <project root>', 'Base dir (root) of your project.')
+  .requiredOption('-b, --basePath <project base path>', 'Base dir path of your project.')
```
egorio, line 58: I think probably -p is better since:

```
tsc -p ./path/to/tsconfig.json
Compiles the TypeScript project located at the specified path.
```

packages/cli/src/commands/migrate.ts
```diff
@@ -232,3 +258,8 @@ function parseModuleList(moduleList: IModule[]): ParsedModuleResult {
   });
   return { edgeList, coreDepList };
 }
+
+// helper function to parse a,b,c to [a, b, c]
+function commaSeparatedArgs(input: string): Array<string> {
```
egorio, line 263: [Seems](https://github.com/rehearsal-js/rehearsal-js/pull/327/files#diff-9c1f10252b43f2fa51543038f9f92b0306f8093823708c92d7d64cb1448ba1dbL10) we prefer to use `string[]`

egorio (approved):

egorio (approved):

lynchbomb (approved):

PR #359, opened 2022-09-09
Use Snapshots and Fixturify Project

chadhietala:
This removes the custom `.input` and `.output` fixture setup in the plugin tests. In general we should be leaning heavily on open source solutions for our testing so that we accrue code that could otherwise be substituted with an open source tool.


lynchbomb (approved):

egorio:
Using `.input` and `.output` concept makes those tests reply on file processing easier to maintain since developer need to create .input file, corresponding .output file and just a function (test) to process the input and compare to output. So - dev always have clear vision of what was input and what is the expected output, not trying to read through the huge test file (for some test we need to have several input files) and trying to "calculate" what is an appropriate snapshot in this case (especially if snap file have several exports).

Here is another [approach](https://github.com/airbnb/ts-migrate/blob/master/packages/ts-migrate-plugins/tests/src/ts-ignore.test.ts) used by ts-migrate team, which is also bit messy :(

We technically had something like `Project.fromDir(root, opts)` which was removed by this PR... probably had to wrap to a class like project, and just need to [extend](https://jestjs.io/docs/expect#expectextendmatchers) `expect` with something  like toMatchFiles(), so test can look like this:

```
{
   project.baseDir = './base_dir';
   project.fromDir('./fixtures_inputs');

   do_something_with('./fixtures');
   
   expect('./base_dir').toMatchFiles('./fixtures_outputs');
   
   project.dispose();
}
```

egorio:
Why `.input` `.output` but not `/input/*.ts` `/output/*.ts` is that .input and .output content are ignored by IDEs and devs can put any content in any formatting and be sure it won't be annoying or fixed by IDE.

Issue #294, opened 2022-08-12
Pass TS Language Service instead of RehearsalService to the codefixes


egorio:
Codefixes don't need to have access to RehearsalService, the only TS LS is needed to gather all required information (program or type checker). 

Codefixes tests also need to be upgraded to remove dependency on `upgrade` command - the test LS service and loop over diagnostic messages can be done in test helpers.

egorio:
Close in favor of https://github.com/rehearsal-js/rehearsal-js/issues/453

PR #360, opened 2022-09-09
chore: only publish the cli

lynchbomb:
As we get closer to GA we want to reduce the public API surface area. This PR flags all packages as `private` except the `cli`.

`Lerna will never publish packages which are marked as private ("private": true in the package.json).`

### DIFF
--- a/packages/codefixes/package.json
+++ b/packages/codefixes/package.json
@@ -1,4 +1,5 @@
 {
+  "private": true,
   "name": "@rehearsal/codefixes",
   "version": "0.0.33",
   "description": "Rehearsal Dependency Codefixes Collection",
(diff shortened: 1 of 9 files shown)


chadhietala:
I don't think its that easy because the CLI currently doesn't bundle so [these deps](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/package.json#L38-L40) are going to error. I have some ideas and cycles if you want me to take this on.

chadhietala (changes requested):
see comment above

lynchbomb:
Yeah we will also need to add rollup etc. for the bundling. Ill just add it to this PR and we can iterate on the config

egorio:
Some of those packages can be useful for community outside of @rehearsal... utils or codefixes (especially after [#294](https://github.com/rehearsal-js/rehearsal-js/issues/294)) can be used separately

PR #361, opened 2022-09-10
bring back fixture files

chadhietala:
This brings back fixture input files to make sure we can control the whitespace of the input. We still should retain the snapshots because they will fail if we mess up the whitespace and its easy to update if the change is intentional.


egorio:
Can we make snapshots like this:
```
exports[`Test EmptyLinesPreservePlugin > run 1`] = readFileSync('file.ts.output');?
```

egorio (approved):

PR #377, opened 2022-09-20
chore: create ember fixtures for testing against an app or addon

wondersloth:
Using scenario-tester, created some variants of an ember app to exercise our use cases for a in-repo-addon and in-repo-engine.


packages/cli/test/commands/upgrade.test.ts
```diff
@@ -16,6 +16,8 @@ const TEST_TSC_VERSION = '4.5.5';
 const ORIGIN_TSC_VERSION = packageJson.devDependencies.typescript;
 let WORKING_BRANCH = '';
 
+const TEST_TIMEOUT = 100000;
```
lynchbomb, line 19: lets remove this line and just up the global test timeout to 100,000
https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/vitest.config.ts#L6

packages/migration-graph/tests/fixtures/ember/addon-template/.editorconfig
```diff
@@ -0,0 +1,19 @@
+# EditorConfig helps developers define and maintain consistent
```
lynchbomb, line 1: can we add all of these config files to the git ignore? do we actually need them to test against for the migration-graph? (i would assume no)

lynchbomb (changes requested):
Why do we need to check all the config files and non app code in? Can we not just add them all to the git ignore? eg. .gitkeep, *.md, .watchmanconfig, *.yml etc.

packages/cli/test/commands/upgrade.test.ts
wondersloth: > Why do we need to check all the config files and non app code in? Can we not just add them all to the git ignore? eg. .gitkeep, *.md, .watchmanconfig, *.yml etc.

Primarily, I wanted to keep the full app functional so I can invoke `ember <build, serve,test>` etc.

I wanted to make it a `ember new <app-name>` that way it's easier to regenerate and replace if needed. I will trim it down for or provide a regenerate script and prune script. So we can recreate this fixtures.

wondersloth: Removed the files, wasn't sure how critical they were. Pruned 👍

wondersloth:
CI is failing because of a failure in  **upgrade.test.ts** in the **cli** package. This fails on master as well on my local.

wondersloth:
I have a separate PR for the linting errors I fixed when I rebased. #385

lynchbomb (approved):

Issue #383, opened 2022-09-23
bug: The same error reported as "fixed: true" and "fixed: false" in the report


egorio:
Found during work on GitHub Action: https://github.com/rehearsal-js/action/pull/21/files

PR #388, opened 2022-09-26
fix: bug in string comparison

brendaHuang:
Fixes this bug: https://github.com/rehearsal-js/rehearsal-js/issues/383


lynchbomb (approved):

egorio:
Thanks, Brenda!

PR #391, opened 2022-09-27
chore: improvements for migrate

haochuan:
This PR:
- In `migrate cli`, use `context` to store all data so that we don't need to call `getMigrationStrategy` multiple times
- Expose `sourceType` in `MigrationStrategy`, so we have a way to tell the project type in the `interactive` mode

### DIFF
--- a/packages/cli/src/commands/migrate.ts
+++ b/packages/cli/src/commands/migrate.ts
@@ -9,6 +9,7 @@ import { Command } from 'commander';
 import { existsSync, readJSONSync, rmSync, writeJsonSync } from 'fs-extra';
 import { Listr } from 'listr2';
 import { resolve } from 'path';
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/migrate.ts
```diff
@@ -57,20 +57,36 @@ migrateCommand
...
 
-    // get custom config
-    const userConfig = options.userConfig
-      ? new UserConfig(options.basePath, options.userConfig, 'migrate')
-      : undefined;
-
     const tasks = new Listr<MigrateCommandContext>(
       [
+        {
+          title: 'Initializatoin',
```
chadhietala, line 63:
```suggestion
          title: 'Initialization',
```

chadhietala (commented):
LGTM

Issue #390, opened 2022-09-27
feat: follow fully qualified service injections for Ember apps


chadhietala:
In Ember apps you can have fully qualified service injections like so:

```js
class SomeComponent extends Component {
  @service('authentication@authenticated-user') authenticatedUser;
}
```

```
'authentication@authenticated-user'
 ~~~~~~~~~~~~~~ ^^^^^^^^^^^^^^^^^^
 ~ addon name   ^service name
```

In these cases we likely want to treat this code as if it was consuming the service via an import. For example, this is what the actual TS code would look like *after* conversion:

```ts
import AuthenticatedUser from 'authentication/services/authenticated-user';

class SomeComponent extends Component {
  @service('authentication@authenticated-user')
  declare authenticatedUser: AuthenticatedUser;
}
```

This would mean that we would add the `authenticated-user service` to the graph order (instead of implying on heuristics) and it should come before `SomeComponent`. 

If we see a service injection with a non qualified injection path

```js
class SomeComponent extends Component {
  @service('tracking') tracking;
  @service('my-service') myService;
}
```

We have to follow this resolution logic:
- checking to see if it's locally defined e.g. `/app/services/*.js`
- If it is there than we need to add it to the graph as we did above.
- If it's not there its coming from some addon. 

In the case where the service is unknown likely going to want to ask where to resolve the service in interactive mode and/or provide a way to give the users a "unresolvable tokens" manifest that they can file out and re-run the migration with to resolve those identifiers.

wondersloth:
I chatted with @chriskrycho about this today. 

I hacked out some simple heuristics based off of file path, and that definitely won't cut it given that services can be used in multiple places components, adapters, routes, services.

The current work I've done from that hack I can carry forward into addressing this feature.

lynchbomb:
In addition to migration ordering, lets keep this issue open as we will also need to handle type inference to handle the import and declaration of the service eg:

```
>>>>> import AuthenticatedUser from 'authentication/services/authenticated-user';

class SomeComponent extends Component {
       @service('authentication@authenticated-user')
>>>>>  declare authenticatedUser: AuthenticatedUser;
}
```

wondersloth:
@lynchbomb 

Just to clarify the additional feature request:

rehearsal should use the meta data from the migration graph to transform the resolver based calls for service decorators, from:

```
import Component from '@glimmer/component';
import { inject as service } from '@ember/service';
import { action } from '@ember/object';

[...]

Issue #365, opened 2022-09-12
Refactor usage of babel in migration-graph-ember to use swc


wondersloth:
In `migration-graph-ember` babel is used when parsing package `main.js` for addons and ember apps.

PR #403, opened 2022-09-29
feat: discover ember service dependency information from a file

wondersloth:
This is work to support feature #390.
This PR addresses parsing a single source file and finding the usages of `@ember/service`. 
It returns an array of service meta data.

```
type EmberInferredServiceDependency = {
  addonName?: string;
  serviceName: string;
};
```

This implementation should not try resolve the services just parse and extract meaningful meta data.


lynchbomb:
Since this implementation is specific to Ember shouldn't we have it live in `migration-graph-ember` and not `migration-graph`?

wondersloth:
> Since this implementation is specific to Ember shouldn't we have it live in `migration-graph-ember` and not `migration-graph`?

- [ ] Move to `migration-graph-ember`

I can move it over there now. My other work is sitting in `migration-graph` I hadn't thought about organization yet because I don't like having to "prepare" across packages.

packages/migration-graph-ember/src/utils/discover-ember-service-dependencies.ts
```diff
@@ -0,0 +1,224 @@
...
+  const maybeImportDeclaration = findImportDeclarationsBySource(parsed.body, '@ember/service');
+
+  if (!maybeImportDeclaration) {
+    DEBUG_CALLBACK('No import declaration found');
+    return EMPTY_RESULT;
+  }
+
+  const importDeclaration = maybeImportDeclaration as ImportDeclaration;
+
+  // In the case of re-assignment from ineject as service, we walk the specifiers
```
chadhietala, line 46:
```suggestion
In the case of re-assignment from inject as service, we walk the specifiers
```

packages/migration-graph-ember/src/utils/discover-ember-service-dependencies.ts
```diff
@@ -0,0 +1,224 @@
...
+    // Check if the ExportDefaultDeclaration has a ClassExpression
+    if (edd.decl.type === 'ClassExpression') {
+      foundClasses.push(edd.decl as ClassExpression);
+    } else {
+      DEBUG_CALLBACK('Default export does not contain a ClassExpression');
+    }
+  }
+
+  if (foundClasses.length < 1) {
+    throw new Error('Unexpected Error: No classes found despite having imported @ember/service.');
```
chadhietala, line 84: I don't think we want to throw in the case. Probably should just return out with an empty result.

chadhietala (changes requested):
Overall looks good to me. Just a couple things that I think need to be fixed.

chadhietala:
Please make sure to migrate the ember graph stuff to use SWC instead of babel

lynchbomb:
https://github.com/rehearsal-js/rehearsal-js/issues/365

lynchbomb (approved):

PR #404, opened 2022-09-29
feat: add rc to cli args

lynchbomb:
This PR moves the cli upgrade to use test snapshots and adds the feature flag of typescript build `rc` to the list and tests against `rc`, `beta`, `latest`


packages/cli/src/commands/upgrade.ts
```diff
@@ -46,7 +46,7 @@ upgradeCommand
   .name('upgrade')
   .description('Upgrade typescript dev-dependency with compilation insights and auto-fix options')
   .argument('[basePath]', 'Path to directory contains tsconfig.json', '.')
-  .option('-b, --build <beta|next|latest>', 'typescript build variant', 'beta')
+  .option('-b, --build <beta|next|latest|rc>', 'typescript build variant', 'beta')
```
egorio, line 49: We also need to add some artificial names for `latest of latest/beta/rc` and `latest of latest/beta/rc/dev` (currently beta-4.9.2, dev-4.9.1 ;) )

Just realized... we can specify them as comma separated list and use biggest number available version.

packages/cli/test/commands/upgrade.test.ts
```diff
@@ -51,44 +53,66 @@ describe('upgrade:command', async () => {
...
-    expect(report.summary.files).toBe(3);
-    expect(report.summary.filesList).toStrictEqual([
-      '/foo/foo.ts',
-      '/foo_2/foo_2a.ts',
-      '/foo_2/foo_2b.ts',
+    expect(report).toMatchSnapshot();
+  });
+});
+
+describe('upgrade:command typescript@rc', async () => {
```
egorio, line 60:
```
describe.each(['rc', 'latest', 'beta'])('upgrade:command typescript@%i', async (buildTag) => {
   ...
}
```

or

```
describe('upgrade:command typescript', async () => {
   test.each(['rc', 'latest', 'beta'])('%1 build`, async (buildTag) => {
      ...
   })
}
```

packages/cli/test/commands/upgrade.test.ts
```diff
@@ -51,44 +53,66 @@ describe('upgrade:command', async () => {
...
+
+    const result = await runTSNode('upgrade', [
+      FIXTURE_APP_PATH,
+      '--report',
+      'json',
+      '--dryRun',
+      '--build',
+      buildTag,
+    ]);
+    // 4.8.4
```
egorio, line 105: Comment will retire soon :)

packages/cli/test/commands/upgrade.test.ts
```diff
@@ -51,44 +53,66 @@ describe('upgrade:command', async () => {
...
-      '2322': 0,
-      '2616': 0,
-      '6133': 0,
-    });
-    expect(report.summary.files).toBe(3);
-    expect(report.summary.filesList).toStrictEqual([
-      '/foo/foo.ts',
-      '/foo_2/foo_2a.ts',
-      '/foo_2/foo_2b.ts',
+    expect(report).toMatchSnapshot();
```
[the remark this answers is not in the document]
lynchbomb: yeah i noticed those dynamic fields after i pushed this PR. i with snapshots had dynamic field annotations.

packages/cli/test/commands/upgrade.test.ts
lynchbomb: i intentionally didn't implement this within a loop so we could have the option to have different checks and conditions for each build. eg. with `rc` we dont want codemods to run and only report etc.

packages/cli/test/commands/upgrade.test.ts
lynchbomb: the comment is just an example of the response
eg. `// 4.8.4`

packages/cli/src/commands/upgrade.ts
lynchbomb: I think `latest-beta` is the most accurate and should be the default. Otherwise we should fetch whichever tag is explicitly being passed

PR #406, opened 2022-09-30
chore: refactor test fixtures

wondersloth:
- Refactor test fixtures to use `fixturify-project` in favor of scenario-tester.
  - Decouple file and project fixtures from scenarios.
- Remove duplicate scenario test.
- Refactor `migration-graph.test.ts` and `migration-strategy.test.ts` to use new project fixtures.


wondersloth:
I have a local branch where I parse the **index.js** to get the **engine-name**. 

Just to note. I am keeping the scenario tests to validate our fixtures correctness.

wondersloth:
CI is failing because `packages/migration-graph/tests/fixtures/ember/(app|addon)-template/` need a node_modules directory in order for fixturify-project to write files.

packages/migration-graph/tests/fixtures/project.ts
```diff
@@ -0,0 +1,109 @@
+import { Project } from 'fixturify-project';
+import { dirname, join } from 'path';
+import merge from 'lodash.merge';
+import tmp from 'tmp';
+import { execSync } from 'child_process';
+import { existsSync } from 'fs-extra';
+
+tmp.setGracefulCleanup();
```
lynchbomb, line 8: minor - but can we fix the ordering here with all the imports above the invocations. eg `tmp.setGracefulCleanup();` moved below the final import

packages/migration-graph/tests/fixtures/project.ts
```diff
@@ -0,0 +1,109 @@
...
+  getEmberAppWithInRepoAddonFiles,
+  getEmberAddonFiles,
+  getEmberAppWithInRepoEngine,
+} from './files';
+import rimraf from 'rimraf';
+
+// this scenario represents the last Ember 3.x release
+export async function ember3(_project: Project) {}
+
+export function getEmberAppProject(project: Project = emberAppTemplate()): Project {
```
lynchbomb, line 21: I think this is ok for now. But for ease of API and redundancy maybe all the variants of getEmber* can be in the same function with a switch statement around type. eg.

```
export function getEmberFixture(name: string, projectType: 'addon' | 'engine' | 'app') {
  switch (projectType) {}
  // ...
}
```

packages/migration-graph/tests/fixtures/scenarios.test.ts
```diff
@@ -3,7 +3,8 @@ import { beforeAll, beforeEach, describe, expect, test } from 'vitest';
 import { addonScenarios, appScenarios, clean, setup } from './scenarios';
 const TEST_TIMEOUT = 500000;
```
lynchbomb, line 4: does this need to be in the file or can we move it to a global config?

packages/migration-graph/tests/fixtures/scenarios.test.ts
```diff
@@ -3,7 +3,8 @@ import { beforeAll, beforeEach, describe, expect, test } from 'vitest';
...
 
-describe('scenarios', () => {
+// Enable these tests to validate if our scenarios fixtures build a funciton all
+describe.skip('scenarios [internal] validate file fixture projects with acceptance tests', () => {
```
lynchbomb, line 7: why are we skipping this? is this a TODO?

packages/migration-graph/tests/migration-strategy.test.ts
```diff
@@ -1,14 +1,22 @@
...
+  getEmberAppProject,
+  getEmberAppWithInRepoAddonProject,
+  getEmberAppWithInRepoEngineProject,
+  setupProject,
+  testSetup,
+} from './fixtures/project';
+
+tmp.setGracefulCleanup();
+
+const TEST_TIMEOUT = 500000;
```
lynchbomb, line 19: move to global?

lynchbomb (changes requested):

packages/migration-graph/tests/fixtures/project.ts
wondersloth: Made the changes.

At first, I used an enum, then I realized that a type is simpler and just as good. No need for the overhead of an enum. Thanks for the suggestion.

PR #425, opened 2022-10-25
refactor: few cleanups in migrate CLI flag/option

haochuan:
Previous `rehearsal migrate` has a few required flag which seems unnecessary, after this cleanup user could simply run `rehearsal migrate` for most of the use cases.

- Removed most of the `--entrypoint` flag in tests cases expect one specific test for that flag, since `migration-graph` has a default entrypoint `index.js`, and it would be able to figure out all the files for ember stuff. But we still keep it as an option for special use cases.
- Remove required `--basePath`, and assign a default value to `process.cwd()`
- Remove `--strict` to create more strict `tsconfig.json`, since it doesn't make sense to have a hard coded strict version in the codebase. If user wants different `tsconfig.json` rather than the minimum one from the code base, they can use the `userConfig` to pass it.


lynchbomb:
This sounds great and I totally agree with the improved UX. Take a peek at the failing build and just want to get @wondersloth input on this before we merge.

haochuan:
> This sounds great and I totally agree with the improved UX. Take a peek at the failing build and just want to get @wondersloth input on this before we merge.

Found an issue in test where I'm using the previous string, fix pushed

lynchbomb (approved):

PR #426, opened 2022-10-25
feat: Add service discovery to migration-graph

wondersloth:
* Remove and replace "file-dependency-graph" with "PackageDependencyGraph" object-oriented code.
* Move the "resolveOptions" for each type of package we will crawl into the the respected PackageDependencyGraph implementation.
* Currently the `PackageDependencyGraph` code is in `packages/migration-graph`. It should be broken apart into `migration-graph-ember` and `migration-graph-shared` respectively.
* There is a problem with this because our test files are all in `packages/migration-graph` right now.

TODO
* [x] `EmberAppPackageDependencyGraph` Add test TODO for a nested service relationship. 
  ```
   // app uses a service `date` from `some-addon`
   // └── some-addon exposes `date` and consumes a service `time` from `another-addon`
   //     └── another-addon exposes a service `time`.
  ```


haochuan (approved):
👍

wondersloth:
One thing I'd like to follow up with is a test-support package that I can store all the fixtures in. This way I can share test fixtures for a library and an ember between migration and CLI.

wondersloth:
This PR should rebase off of #426 once it lands.

PR #431, opened 2022-10-28
fix: improve service dependency graph for addons

wondersloth:
- fix: bug where an `EmberAddon` was not making associating between it's interface and the implementation. e.g. `app/components/hello.js` and `addon/component/hello.js`.
  - Test added to validate the creation of the edge in the graph.
- fix: Improved the test to validate how an in-repo addon is added to the graph.
- fix: Fixed a bug where when trying to create a graph for a partially migrated ember app we would try and parse the TS file.
  - Current implementation of `discoverEmberServiceDependencies` does not parse `.ts` files.
- chore: rename `EmberPackage` to `EmberAppPackage` to align with naming conventions.
- chore: refactored how the return type for Graph.getNode. 
  - Previously it was returning the `Node | undefined` this was the exact result from the registry (set).
  - Throwing an error at the getNode API simplifies the type checking when getNode is used, typically because we guard with a `hasNode()` statement.
- chore: fix typo in test variable.
- chore: remove unnecessary usages of `lodash.merge` where applicable.


packages/migration-graph-ember/src/utils/discover-ember-service-dependencies.ts
```diff
@@ -32,6 +32,11 @@ export function discoverServiceDependencies(
 ): EmberInferredServiceDependency[] {
   const filePath = join(baseDir, pathToFile);
 
+  // TODO Evaluate if we need service discovery of TS files. Currently this parser doesn't work for typescript source.
+  if (extname(filePath) !== '.js') {
```
lynchbomb, line 36: This should be fine as we can manage state of .ts files in the cli/migrate

lynchbomb (approved):

packages/migration-graph-ember/src/utils/discover-ember-service-dependencies.ts
wondersloth: When I tried ac-hub-web it had a partially migrated in-repo addon and 💥 it broke. So this was my hack to get around that.

PR #432, opened 2022-11-01
chore: cleanup of incorrect async types and replace winston logging for debug

lynchbomb:
This PR:
- cleans up incorrect async types for sync operations
- replaces noisy winston logging with debug callbacks eg. `DEBUG=rehearsal:* rehearsal <CMD> <ARGS>`


packages/plugins/src/plugins/diagnostic-fix.plugin.ts
```diff
@@ -21,14 +23,14 @@ export class DiagnosticFixPlugin extends Plugin {
     let diagnostics = this.getDiagnostics(fileName, commentTag);
     let tries = diagnostics.length + 1;
 
-    this.logger?.debug(`Plugin 'DiagnosticAutofix' run on ${fileName}`);
```
egorio, line 24: Why do you replace logger with debug?
Winston is used as CLI output https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/upgrade.ts#L65-L67

packages/cli/src/commands/migrate.ts
```diff
@@ -11,9 +11,10 @@ import {
 } from '@rehearsal/migration-graph';
 import { jsonFormatter, mdFormatter, Reporter, sarifFormatter } from '@rehearsal/reporter';
 import { Command } from 'commander';
-import { existsSync, readJSONSync, writeJsonSync } from 'fs-extra';
+import { existsSync } from 'fs-extra';
```
egorio, line 14: You can use native import { existsSync } from 'fs';

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
lynchbomb: > Why do you replace logger with debug?
winston.logger is noisy and other than actually debugging an issue, in-which we would use the common `DEBUG` module. 

All other logging for Rehearsal should happen within Commander or throw. I can't think of any other uses?

packages/upgrade/src/upgrade.ts
```diff
@@ -42,8 +45,7 @@ export async function upgrade(input: UpgradeInput): Promise<UpgradeOutput> {
     LintPlugin,
   ];
 
-  logger?.info('Upgrade started.');
-  logger?.info(`Base path: ${basePath}`);
```
egorio, line 46: Let's keep the logger at least in this file to show progress, otherwise the user won't have any feedback - if CLI is stuck or working.

lynchbomb: We should propagate the log stream into commander then

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
egorio: The transport layer has a `level` option to set the maximum output level.
We can use `info` as a default level and change to `debug` if `--verbose` option passed.

PR #440, opened 2022-11-06
feat: migration state

haochuan:
This PR adds migration state to keep track of the progress of migration in interactive mode, including:
1. Create a state file `.rehearsal/migration-state.json` and `git add` it
2. Every time when running `rehearsal migration`, it would load the previous state if there is any, verify and update the state if needed.
  - Make sure migration status in state file matches what we have on disk
  - Update ts-ignore count for each migrated file
4. There are 3 cases when showing the migration progress/state in interactive mode:
  - package has not started any migration before -> show "no progress found"
  - package had a migration with remaining JS files ->  show "x/x migrated, x rehearsal todos"
  - fully migrated with no JS and rehearsal todo -> show "completed" and disable this option

Screen Shot 2022-11-06 at 01 43 01

5. We had discussion about making the state file readable to other tools/platforms, I could have follow up PR for that if needed.

### DIFF
--- a/packages/cli/src/commands/migrate.ts
+++ b/packages/cli/src/commands/migrate.ts
@@ -19,11 +19,15 @@ import { generateReports } from '../helpers/report';
 import {
   MigrateCommandContext,
   MigrateCommandOptions,
-  PackageMap,
+  packageChoiceMap,
   PackageSelection,
 } from '../types';
 import { UserConfig } from '../userConfig';
 import { addDep, determineProjectName, parseCommaSeparatedList, runModuleCommand } from '../utils';
+import { State } from '../helpers/state';
+
+const IN_PROGRESS_MARK = '🚧';
+const COMPLETION_MARK = '✅';
 
 function ifHasTypescriptInDevdep(basePath: string): boolean {
   const packageJSONPath = resolve(basePath, 'package.json');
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/migrate.ts
```diff
@@ -74,45 +72,82 @@ migrateCommand
...
-              const packageSelections: PackageSelection[] = discoverEmberPackages(
-                options.basePath
-              ).map((p) => {
-                let { packageName: name } = p;
-                if (name === projectName) {
-                  // a little better message to indicate this option is for entire project
-                  name = `${name} (this entire project)`;
+              // Init state and store
+              const state = new State(
+                projectName as string,
```
lynchbomb, line 80: if we can avoid this type cast lets do it

lynchbomb:
This is great! Follow-up PR(s) if the state class can produce a SARIF file (similar to upgrade / reporter) dev's can leverage an existing workflow they are accustomed eg. Checkup.js. via VSCode SARIF plugin which renders the todo' items as .csv for easy viewing.

And I know we talked about a child command in the CLI for things like "TODO Parsing". eg

`rehearsal migrate status` 
`rehearsal upgrade status`

which will read the state artifact and in terminal show the entire list of DONE/TODOs for the given parent command (migrate or upgrade)

PR #450, opened 2022-11-10
chore: remove babel, improve graph creation performance for large ember apps

wondersloth:
**Added**

- Tests for `ProjectGraph.ts`
- Added `options.eager` to `ProjectGraph` API. Explained below.

**Changed**

- Now, by default a `ProjectGraph` will lazily create a moduleGraph for each
  `addPackageToGraph()` call.
- This is important in the case of a large ember-app:
  - If we eagerly create a package's module graph, we start discovering services.
  - For those discovered services, if they're from a package (in-repo-addon) within
    the project, we may not have the correct stub (synthetic) GraphNode created for that
    package.
  - This may result in packages in the output that are for the ember-addon's moduleName and
    not the correct `package.json` packageName.
  - Creating an unintended side effect, where we have incorrect edges created for package's
    dependency on a service.
- When `options.eager=true`
  - We will create a moduleGraph on every `addPackageToGraph` call.
  - This has some challenges for ember applications.
  - When we discover services, we have the fully qualified ember name, but the graph may only
    have `moduleName` and not have the `packageName`.
  - If a synthetic node is created with only a `moduleName` we currently don't update it since
    we're always looking up by `packageName`.
  - This works now (added in latest push).

- Moved the logic for `emberDiscoverPackages` into `EmberAppProjectGraph.discover()`.

**Removed**

- Removed all babel code from `migration-graph-ember` the APIs were not used.
  - The babel parsing that was being done was for the `moduleMain` or `index.js` in side of an addon.
  - We don't try and parse those files, because we can simple require the `index.js` (main)
    file and ask what the `moduleName` is.
  - Removed tests.

# TODO
- [x] Create test TODO for the `options.eager` for partially discovered service where moduleName doesn't match packageName.


wondersloth:
This PR also addresses #390

wondersloth:
This PR also address #365

lynchbomb (approved):

PR #451, opened 2022-11-14
feat: improve UX when running tsc command

haochuan:
Couple of UX improvements:
- Move running `tsc` to the end of all tasks
- Although we should expect the migration covers all the stuff and `tsc` should exit on `0`, just for rare cases additional message has been added to let the user know what happened.

### DIFF
--- a/packages/cli/src/commands/migrate.ts
+++ b/packages/cli/src/commands/migrate.ts
@@ -14,6 +14,8 @@ import { Command } from 'commander';
 import { existsSync } from 'fs-extra';
 import { Listr } from 'listr2';
 import { createLogger, format, transports } from 'winston';
+import chalk from 'chalk';
+import { createLogger, format, transports } from 'winston';
 import { debug } from 'debug';
 
 import { generateReports } from '../helpers/report';
@@ -26,6 +28,7 @@ import {
   runModuleCommand,
   writeTSConfig,
   isTypescriptInDevdep,
+  getModuleManager,
 } from '../utils';
 import { State } from '../helpers/state';
 
@@ -237,15 +240,6 @@ migrateCommand
           },
         },
         {
-          title: 'Checking for TypeScript errors',
-          enabled: (ctx): boolean => !ctx.skip,
-          task: async () => {
-            await runModuleCommand(['tsc'], { cwd: options.basePath });
-          },
-        },
-        // TODO: what to do with those ts errors?
-
-        {
           title: 'Creating eslint config',
           enabled: (ctx): boolean => !ctx.skip,
           task: async (_ctx, task) => {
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/migrate.ts
```diff
@@ -257,6 +250,29 @@ migrateCommand
...
+            } catch (e) {
+              // workaround for type e = unknown or any
+              let message;
+              if (e instanceof Error) {
+                message = e.message;
+              } else {
+                message = String(e);
+              }
+              const hintText = chalk.yellow(
+                `The migration is completed, but there are some extra TS erros listed above to be fixed manually.\nYou can run '${moduleManager} tsc' to get the errors, or to verify after the fix.`
```
chadhietala, line 270: When does this occur? Shouldn't we have annotated all the places that error with `@ts-expect-error` by the time we exit? Maybe we should have a test for this?

```suggestion
The migration has completed, but there are some extra TS errors listed above to be fixed manually.\nYou can run '${moduleManager} tsc' to get the errors, or to verify after the fix.
```

packages/cli/src/commands/migrate.ts
```diff
@@ -257,6 +250,29 @@ migrateCommand
             }
           },
         },
+        {
+          title: 'Checking for TypeScript errors',
+          enabled: (ctx): boolean => !ctx.skip,
+          task: async (_ctx, task) => {
+            const moduleManager = await getModuleManager();
+            try {
+              await runModuleCommand(['tsc'], { cwd: options.basePath });
```
chadhietala, line 259: Just a note, but we're going to need to figure out how we select between `tsc` and using `glint` here. Probably need to plumb it through `ctx` in the initialization step.

packages/cli/src/commands/migrate.ts
lynchbomb: ^^^^ this. essentially if this is falling through it is a bug. we wouldn't expect a user to have to manually run `tsc` to glean insight into the compiler errors. Rehearsal should have an annotation around every error and associated "task" in the Rehearsal Report.

packages/cli/src/commands/migrate.ts
lynchbomb: @brendaHuang ^^^^

packages/cli/src/commands/migrate.ts
haochuan: @chadhietala @lynchbomb Yeah totally agree on these, and I've mentioned in the description that we should expect the migration cli covers all the stuff and `tsc` command should always exit on 0. However, IMO I feel better to have this extra guard, since I'm not confident to say that end users would never see this at this moment.

And now I'm also thinking since we see this `tsc` failing as a bug, could we just remove this task to run `tsc` command?

lynchbomb: > And now I'm also thinking since we see this tsc failing as a bug, could we just remove this task to run tsc command?

Interesting yeah. Essentially what value add is running tsc after @rehearsal/migrate has already culled all the tsc errors as TODO's into the Rehearsal report. 

Besides having a guard around and missed tsc errors from @rehearsal/migrate, I dont actually think we need to run `tsc` (to your point).

We will def. need a summation log something like eg. Rehearsal has automated `N` codefixes across `N` files. Rehearsal manual type tightening TODO's remaining `N`.

You could think if that remaining tasks is `0` when running `tsc` we would expect this to pass with strict types. So we *could* wrap a conditional around `tsc` and catch if tasks === 0 and tsc is exiting a non 0?

haochuan:
Addressed this one in #462

Issue #448, opened 2022-11-08
@rehearsal/migrate tsconfig strict


lynchbomb:
I think we will need to tighten the tsconfig we are [generating](https://github.com/rehearsal-js/rehearsal-js/blob/f8603fbdf82072a819f07fe9a96e7a99185c67c5/packages/cli/src/utils.ts#L342-L356) for the migrate command to be strict (copy tsBases config for a starter)

That way we are generating MORE rehearsal todo tasks, since at the end of the day we want apps to migrate to be as strict as possible.

PR #452, opened 2022-11-14
feat: tighten default tsconfig

haochuan:
For #448

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -352,6 +352,7 @@ export function writeTSConfig(basePath: string, fileList: string[]): void {
       target: 'ES2021',
       forceConsistentCasingInFileNames: true,
       noFallthroughCasesInSwitch: true,
+      noEmit: true,
     },
     include,
   };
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils.ts
```diff
@@ -342,15 +342,16 @@ export function parseCommaSeparatedList(value: string): string[] {
...
-      target: 'es2016',
-      module: 'commonjs',
+      strict: true,
+      noUncheckedIndexedAccess: true,
+      module: 'es2020',
       moduleResolution: 'node',
+      newLine: 'lf',
+      target: 'ES2021',
+      forceConsistentCasingInFileNames: true,
+      noFallthroughCasesInSwitch: true,
```
chadhietala, line 354: Don't we want `"noEmit": true` to be in here since we are removing the `outDir`?

lynchbomb: @chadhietala yes. depending on type inference work we might need to generate the d.ts files. but yeah as far as i can see we do not want the compiler output files.

haochuan: 👍

lynchbomb (approved):

PR #454, opened 2022-11-15
fix: Transforms change source code at wrong positions

egorio:
Not ready


packages/codefixes/package.json
```diff
@@ -35,6 +35,7 @@
     "@rehearsal/utils": "workspace:*"
   },
   "devDependencies": {
+    "@rehearsal/upgrade": "workspace:*",
```
lynchbomb, line 38: isn't this a circular dependency?

lynchbomb (changes requested):

packages/codefixes/package.json
chadhietala: Maybe I'm missing something but it doesn't seemed used.

egorio: :(

egorio: Moved test back to `upgrade`... will update test later after CodeFixAction

lynchbomb (approved):

PR #458, opened 2022-11-22
chore: refactor reporter

chadhietala:
WIP

This is an attempt to refactor the reporter to just produce Sarif format directly instead of an intermediate representation that is then converted into Sarif via a formatter.

### DIFF
--- a/packages/reporter/src/formatters/md-formatter.ts
+++ b/packages/reporter/src/formatters/md-formatter.ts
@@ -1,32 +1,38 @@
 import { Report } from '../types';
 
 export function mdFormatter(report: Report): string {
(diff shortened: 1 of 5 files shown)


egorio:
Why use serif directly?
Why didn't we discuss this before changing? 
The idea is to internally store info in some simple (as raw as possible) format and convert it to any others.
I'm planning to get rid of FixTransforms in favor of ts.CodeFixActions (after [this](https://github.com/rehearsal-js/rehearsal-js/pull/459/files)) and send them directly to the reporter to collect changes that were applied to files. Then we can convert them to anything else.

Issue #424, opened 2022-10-25
Inherited Object Types


chadhietala:
Note: This should be attempted only once we feel good about library code getting transformed.

One of the challenges for migration is going to be dealing with framework specific objects. This is challenging because TypeScript offers no code action for filling in types from the super class.

[Example](https://www.typescriptlang.org/play?#code/MYGwhgzhAECCB2BLAtmE0DeAoavoTAE8AKZAUyjAHMyAufAFwCdF4qBKeiZ1qzHPIKZkGAVybxoAcgAWZECAD2UgNwDcAXyxasoSDAAiivmQAeDMvAAmMBCjT9BBEuUo12maAHov0AMKKVmRwwAyIipJywtAA7ogK0GISiTKIMKwMitAABq4ENPRg8ITZ6oLQwkmSUgBGYExSZVo6elD+YAzQZhbWtkio6NiCigBuZEwsQfhEpBT5ZB4Y3r4BU7Ch4ZHjwXEJVSlp0BlZuXPUdNBFJZcwMfIgZUIi4tV1DU3aQA)

Since methods on framework objects typically get called by the framework it self instead of other application code the signatures are actually well known and should not vary. Given this aspect I think we can get very far at emitting the correct framework types by transforming the code in 3 stages to get code actions to do much of the heavy lifting.

Stage 1:

In stage 1 we exam the code and determine if a framework object is being extended. We then look at that type and determine what what members it contains. From there we look at what the developer overrode and we build a temporary type that `Pick`s the values out of the super class type.

[...]

PR #459, opened 2022-11-28
feat: use TypeScript's codefixes (including type inference)

egorio:
WIP


packages/codefixes/src/hints-provider.ts
```diff
@@ -1,56 +1,21 @@
...
-};
+import { CodeHintList, DiagnosticWithContext } from './types';
 
 /**
- * Provides
+ * Provides access to useful hints for Diagnostics
  */
-export class CodeFixCollection {
-  readonly list: CodefixList;
+export class HintsProvider {
```
egorio, line 12: Thinking about moving hints to a separate package...

packages/codefixes/src/codefixes-provider.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  constructor(collections: CodeFixCollection[]) {
+    this.collections = collections;
+  }
+
+  /**
+   * Returns a list of FixTransforms.
+   */
+  getCodeFixes(diagnostic: DiagnosticWithContext): FixTransform[] {
+    const fixes: FixTransform[] = [];
```
egorio, line 17: Planning to replace FixTransform with ts.CodeFixAction-kind formatted fixes.

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
```diff
@@ -34,13 +34,19 @@ export class DiagnosticFixPlugin extends Plugin {
...
 
-      const fix = codefixes.getFixForDiagnostic(diagnostic);
-      const fixedFiles: FixedFile[] = fix?.run(diagnostic, this.service) || [];
-      const fixed = fixedFiles.length > 0;
+      let fixedFiles: FixedFile[] = [];
+
+      const fixes = codefixes.getCodeFixes(diagnostic);
+
+      while (fixes.length > 0 && fixedFiles.length === 0) {
+        fixedFiles = fixes.shift()?.run(diagnostic, this.service) || [];
```
egorio, line 42: This is considered a hack because we currently can't predict if FixTransform will be applied or not, so we have to try to apply a fix and check the result to know if it was used. This code will be cleaned out after Transforms is replaced with CodeFixActions in the next iteration, so getCodeFixes will be returned only one CodeFixAction.

packages/upgrade/src/upgrade.ts
```diff
@@ -1,11 +1,5 @@
 import { dirname, resolve } from 'path';
-import {
-  DiagnosticFixPlugin,
-  EmptyLinesPreservePlugin,
-  EmptyLinesRestorePlugin,
-  LintPlugin,
-  ReRehearsePlugin,
-} from '@rehearsal/plugins';
+import { DiagnosticFixPlugin, LintPlugin, ReRehearsePlugin } from '@rehearsal/plugins';
```
egorio, line 2: We don't need EmptyLinesPreservePlugin and EmptyLinesRestorePlugin anymore! Will be cleaned up after more tests.

packages/upgrade/test/fixtures/upgrade/ts-errors.ts.output
```diff
@@ -35,32 +22,13 @@ export function usesMissingVarWithLongName(): boolean {
...
-    const b = null;
-    /* @ts-ignore @rehearsal TODO TS2322: The function expects to return 'void', but 'string' is returned. Please convert 'string' value to 'void' or update the function's return type. */
-    return 'a';
-  }
+// The next function call is not properly formatted
+timestamp();
 
-  return resolved;
+function missingFunction(): void {
+  throw new Error('Function not implemented.');
```
egorio, line 29: This is one of TS CodeFixes behavior... we can decide override this behavior by removing function (or add comment as before)

packages/codefixes/src/7006/index.ts
```diff
@@ -1,91 +0,0 @@
-import { findNodeAtPosition, insertIntoText } from '@rehearsal/utils';
-import { getDefaultFormatCodeSettings } from 'typescript';
-
-import { type FixedFile, FixTransform } from '../fix-transform';
-import type { DiagnosticWithLocation, TextChange } from 'typescript';
-import type { RehearsalService } from '@rehearsal/service';
-
-export class FixTransform7006 extends FixTransform {
```
egorio, line 8: This is the same as 6133 covered by TS

packages/codefixes/src/7006/index.ts
```diff
@@ -1,91 +0,0 @@
...
-      formatOptions,
-      {}
-    );
-
-    const stubLocation = { startLine: 1, startColumn: 1, endLine: 1, endColumn: 1 }; // TODO: actually get the correct coordinates for this.
-
-    fixes.forEach((fix) => {
-      if (fix.fixId) {
-        // Why do we need a combined change?
-        const combined = languageService.getCombinedCodeFix(
```
egorio, line 38: Tripple-checked :)
We don't need to use getCombinedCodeFix in our implementation of our plugin where we do fix error by error.
 
 getCombinedCodeFix returns fixes of the same kind per file... in IDE it is called "Fix all ..."

packages/upgrade/test/fixtures/upgrade/ts-types.ts.output
```diff
@@ -0,0 +1,3 @@
+export function test(me: string) {
```
egorio, line 1: Example of type inference.

lynchbomb:
This is great. Going forward we will need to validate the priority order (jsDoc, TS, Rehearsal) of codemods based on the number of Rehearsal TODO's on actually repo's. eg in the follow codemod priority order:

A / B / C = 26 TODO's
B / C / A = 33 TODO's
C / A / B = 10 TODO's

- once we are in a good place with this; be sure we are covering eg [framework-object-types](https://github.com/rehearsal-js/rehearsal-js/issues/424)
- and handle the integration of consuming superset ts language service eg. glint

lynchbomb (approved):

PR #462, opened 2022-11-28
feat: adding migration summary

haochuan:
As we discussed in #451, running `tsc` after the JS->TS conversion seems unnecessary, since we are expecting to catch all the errors via the CLI. This PR removes `tsc` command, and adding a summary message after the actual conversion like:

```
✔ 30 JS files has been converted to TS, and there are 552 errors caught by rehearsal
  - 3 have been fixed automatically by rehearsal
  - 549 have been updated with @ts-ignore @rehearsal TODO which need further manual check
```

This PR also update the report flag, to have `json` report by default, and update the report formatter in `migrate`, to match the same in `upgrade` command.


packages/cli/src/commands/migrate.ts
```diff
@@ -48,7 +47,7 @@ migrateCommand
     '-r, --report <reportTypes>',
     'Report types separated by comma, e.g. -r json,sarif,md',
     parseCommaSeparatedList,
-    []
+    ['json']
```
lynchbomb, line 50: we want the default to be sarif

packages/cli/test/commands/migrate.test.ts
```diff
@@ -216,15 +226,15 @@ describe('migrate - handle custom basePath', async () => {
 
   test('Run cli againt specific basePath via -basePath option', async () => {
     const customBasePath = resolve(basePath, 'base');
-    const result = await runBin('migrate', ['--basePath', customBasePath], {
+    const result = await runBin('migrate', ['--basePath', customBasePath, '-v'], {
       cwd: basePath, // still run cli in basePath
```
lynchbomb, line 230: `cwd: basePath` is causing failures ... if you rebase this will/should be removed

lynchbomb (changes requested):

haochuan:
Resolved overlaps and conflict with current master

lynchbomb (approved):

Issue #461, opened 2022-11-28
Codemod against @tracked property


lynchbomb:
After running @rehearsal/cli#migrate from latest master. Looks like an issue mutating on codemod with `@tracked` properties. Because this mutating ends with malformed javascript this would cause a build/test failure; and prevents rehearsal from properly inferring types from the remainder of the file.

**input:**
``` ts
export default class FooService extends Service {
  /**
   * Current viewport size
   * @type {string}
   */
  @tracked size = 'l';

  /**
   * Current viewport with in pixels
   * @type {string}
   */
  @tracked viewportWidth;

  /**
   * Whether the viewport is equal to the given size
   * @param {string} size
   * @returns {boolean}
   */
  isBreakpoint(size) {
    return this.size === size;
  }
}
```

**output:**
``` ts
export default class FooService extends Service {
  /**
   * Current viewport size
   * @type {string}
   */
  @tracked size = 'l';

  /**
   * Current viewport with in pixels
   * @type {string}
   */
  @tracked
/* @ts-ignore @rehearsal TODO TS2532: Object is possibly 'undefined'. */
    viewportWidth!: number;| undefined;

  /**
   * Whether the viewport is equal to the given size
   * @param {string} size
   * @returns {boolean}
   */
  breakpoints(size) {
    return this.size === size;
  }
}
```

egorio:
Seems similar to https://github.com/rehearsal-js/rehearsal-js/pull/454

egorio:
Seems fixed.

![image](https://user-images.githubusercontent.com/1502496/205403657-ddf160ea-2aec-4c4b-bda8-086e1cd53aee.png)

Issue #460, opened 2022-11-28
Codemod excessive inline comments


lynchbomb:
After running @rehearsal/cli#migrate from latest master, against a given project the following input/output results. Note the output is trimmed in the example below, however the actual file the "@ts-ignore @rehearsal ..." comments are repeated over 8000 times:

**input:**

``` ts
export default class Foo extends Service {
  constructor() {
    super(...arguments);
  }

  fooMethod(size) {
    const nextBreakpointIndex =
      orderedBreakpoints.indexOf(
        orderedBreakpoints.find(([key]) => key === size)
      ) + 1;
    if (nextBreakpointIndex < orderedBreakpoints.length) {
      return this.viewportWidth >= orderedBreakpoints[nextBreakpointIndex][1];
    }
    // If this point is reached, searching for the max breakpoint
    return false;
  }
}
```

**output:**

``` ts
export default class Foo extends Service {
  constructor() {
    super(...arguments);
  }

  fooMethod(size) {
    const nextBreakpointIndex =
      orderedBreakpoints.indexOf(
/* @ts-ignore @rehearsal TODO TS2345: Argument of type 'string, unknown] | undefined' is not assignable to parameter of type 'string, unknown'. Consider verifying both types, using type assertion: '(
        orderedBreakpoints.find(([key]) => key === size) as string)', or using type guard: 'if (
        orderedBreakpoints.find(([key]) => key === size) instanceof string) { ... }'. */
/* @ts-ignore @rehearsal TODO TS2345: Argument of type 'string, unknown] | undefined' is not assignable to parameter of type 'string, unknown'. Consider verifying both types, using type assertion: '(
/* @ts-ignore @rehearsal TODO TS2345: Argument of type 'string, unknown] | undefined' is not assignable to parameter of type 'string, unknown'. Consider verifying both types, using type assertion: '(
        orderedBreakpoints.find(([key]) => key === size) as string)', or using type guard: 'if (
        orderedBreakpoints.find(([key]) => key === size) instanceof string) { ... }'. */
/* @ts-ignore @rehearsal TODO TS2345: Argument of type 'string' is not assignable to parameter of type 'string, unknown'. Consider verifying both types, using type assertion: '(
/* @ts-ignore @rehearsal TODO TS2345: Argument of type 'string, unknown] | undefined' is not assignable to parameter of type 'string, unknown'. Consider verifying both types, using type assertion: '(
[...]

Issue #453, opened 2022-11-14
Codefies to use DiagnosticWithContext


egorio:
We already preparing DiagnosticWithContext which includes affected node, type-checker etc. in the diagnostic-fix.plugin, so we don't need to do:

```
const node = findNodeAtPosition(diagnostic.file, diagnostic.start, diagnostic.length);
const program = service.getLanguageService().getProgram()!;
const checker = program.getTypeChecker();
```

inside codefixes. 

Need to update the codefixes to use DiagnosticWithContext instead of DiagnosticWithLocation.

Issue #312, opened 2022-08-25
Refactor CodeFixes to not rely on module scope state


egorio:
Do similar to https://github.com/rehearsal-js/rehearsal-js/pull/311 for CodeFixes

PR #465, opened 2022-11-29
refactor: use TextChanges for our fixes implementation

egorio:
- Replace FixTransfofrms with native CodeFixAction
- Move hints to HintCollection to apply them the same way as other fixes
- Add `newCode` and `oldCode` props to the report (https://github.com/rehearsal-js/rehearsal-js/pull/471) 
- Remove RehearsalService dependency from codefixes
- Codefixes now uses DiagnosticWithContext 
- Add SuggestedDiagnostics to utilize JSDoc type inference

Tests will updated later. 
It will be way easier to debug issues like [this](https://github.com/rehearsal-js/rehearsal-js/issues/461) and [this](https://github.com/rehearsal-js/rehearsal-js/issues/460) because we need to check CodeFixAction content to see what is expected to be changed in the file :)

Close:
- #453
- #312
- #310


packages/upgrade/test/fixtures/upgrade/ts-types.ts.output
```diff
@@ -1,3 +1,39 @@
...
+/**
+ * JSDoc is ignored
+ * @param {number} me
+ * @return {string}
+ */
+export function test3(me: number): string {
+  return me.toString();
+}
+
+/* @ts-ignore @rehearsal TODO TS2345: Argument of type 'string' is not assignable to parameter of type 'number'. Consider verifying both types, using type assertion: '('O' as string)', or using type guard: 'if ('O' instanceof string) { ... }'. */
```
egorio, line 27: This is an interesting example. Types come from JSDoc (number) conflicts with type from usage (string)

chadhietala:
😄

lynchbomb (approved):

PR #466, opened 2022-11-29
fix: use json5 to parse

lynchbomb:
This resolves an issue we could have when comments are contained within tsconfig.json. We should use JSON5 instead of readJSON to handle.


packages/cli/src/types.ts
```diff
@@ -75,3 +75,9 @@ export type PackageSelection = {
 export type MenuMap = {
   [key: string]: string;
 };
+
+export type TSConfig = {
+  compilerOptions: {
```
egorio, line 80: ts.CompilerOptions

PR #471, opened 2022-12-01
refactor: replace code with newCode and oldCode

brendaHuang:
@egorio Replacing `code` field with `newCode` and `oldCode`.


egorio:
Could we merge this after https://github.com/rehearsal-js/rehearsal-js/pull/465 please :)

Issue #463, opened 2022-11-29
migrate cli with ember app/addon


lynchbomb:
In the use case of migrating a given Ember App or Addon from JS to TS. During the [Installing Dependencies](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/migrate.ts#L169-L184) task, a conditional check for eg. `isEmber` we should also install all [pre-req's](https://typed-ember.gitbook.io/glint/using-glint/ember/installation#version-requirements) for Glint and Ember TypeScript (https://docs.ember-cli-typescript.com/installation).

Install Pre-Req Deps for Glint w/Ember Loose and Ember TypeScript:

```
yarn add --dev @glint/core @glint/template @glint/environment-ember-loose

ember install ember-cli-typescript@latest
```

^^^^ By doing above before executing [migrate](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/migrate.ts#L220) the quality of type inference will be significantly better.

lynchbomb:
We are going to implement this via "rehearsal-config.json" and expose an optional validator function hook. That way we can keep framework specific code as a plugin vs core feature:

https://github.com/rehearsal-js/rehearsal-js/pull/477#issuecomment-1338130078

haochuan:
This is handled and completed internally via `rehearsal-config.json`

PR #477, opened 2022-12-01
feat: install extra deps for ember project

haochuan:
PR for #463, CLI would hard stop and show the error before if running migration on an Ember project that doesn't meet the requirements:

```
rehearsal migrate
✖ Your Ember project does not meet the requirements of migration. Please make sure you have:
  - ✅ Project Node Version v16.18.1 should >= 16.0.0
  - ❌ Dependency ember-source ^3.10.0 should >=3.24.0
  - ❌ Dependency ember-modifier 3.1.7 should >=3.2.7
◼ Initialization
◼ Installing dependencies
◼ Creating tsconfig.json
◼ Converting JS files to TS
◼ Creating eslint config
```

Also `exitOnError` has been changed to `true` in `rehearsal migrate`, in order to hard stop the CLI if there is any issue. The reason we had it on `false` was that we don't want to quit the CLI if there we got non 0 exit during `tsc` check. This would also cover #481.


packages/cli/src/commands/migrate.ts
```diff
@@ -182,13 +183,32 @@ migrateCommand
...
               await _ctx.userConfig.install();
             }
+
             // even if typescript is installed, exec this and get the latest patch
             await addDep(['typescript'], true, { cwd: options.basePath });
+
+            // extra dependencies for Ember App/Addon/Engine
+            // TODO: dependes on how much extra stuff we need for a specific framework,
+            // probably need a plugable system for this.
+            if (isEmber(packageJson)) {
```
lynchbomb, line 199: we will need a conditional around which version of Ember and which version of Node. 

Glint only supports:
```
Ember 3.24+
Node 16+
```

lynchbomb (changes requested):

packages/cli/src/commands/migrate.ts
lynchbomb: ... and 

`ember-modifier 3.2.7+`

egorio: Weird to see this as a part of migrate.ts. Isn't better to move this part to a framework-specific package/function (like we did with migration --grapth / migration-graph-ember)?

Move this `['@glint/core', '@glint/template', '@glint/environment-ember-loose']` outside of migrate.ts (params?) and just check `if (additionalDependencies) { ... install them ...}`?

haochuan: @egorio yeah totally agree, that's why in the comment I have the TODO to take it out of the main cli and make it like a plugin-ish structure, still deciding where those should be and I'm open to any suggestions 😄

egorio: > probably just move it to a new file

Oh, didn't read a comment :) I think it's fine to put in a separate file for now :)

haochuan:
Having a wired issue during `./node_modules/.bin/ember install ember-cli-typescipt@latest`, which failed the test:

```
Command failed with exit code 1: npm install --save-dev --loglevel error ember-cli-typescript@latest
npm ERR! Cannot read properties of null (reading 'name')
```

lynchbomb:
@haochuan i've not debugged that comment above. My knee-jerk is `npm install` <<< if this is trying to run on this repo we are using `pnpm` and `npm` execs will throw.

haochuan:
> @haochuan i've not debugged that comment above. My knee-jerk is `npm install` <<< if this is trying to run on this repo we are using `pnpm` and `npm` execs will throw.

Yeah seems like all the utils function to detect 'moduleManager` etc would get the value from `process.cwd()` instead of the actual test tmp dir, I'll update them to see if that would solve the issue

haochuan:
Synced with @lynchbomb  and we both agreed to take any framework specific stuff out of the main CLI. I'm going to close this PR and try to see what I can do to have this ember validator and extra deps installer using the existed `userConfig`

Issue #486, opened 2022-12-05
@rehearsal/migrate package.json#scripts


lynchbomb:
As a migration step we need to include a package.json scripts hooks for build and lint eg:

package.json#scripts

- "build:tsc": "tsc -b"
- "lint:tsc": "tsc --noEmit"

haochuan:
Done in #503

PR #503, opened 2022-12-07
fix: add TS related scripts in package.json

haochuan:
For #486.

### DIFF
--- a/packages/cli/src/commands/migrate.ts
+++ b/packages/cli/src/commands/migrate.ts
@@ -40,6 +40,7 @@ import {
   writeTSConfig,
   getPathToBinary,
   readJSON,
+  addPakcageJsonScripts,
 } from '../utils';
 import { State } from '../helpers/state';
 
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/migrate.ts
```diff
@@ -40,6 +40,7 @@ import {
   writeTSConfig,
   getPathToBinary,
   readJSON,
+  addPakcageJsonScripts,
```
lynchbomb, line 43: typo. should be `addPackageJsonScripts`

packages/cli/src/commands/migrate.ts
```diff
@@ -285,6 +285,16 @@ migrateCommand
             }
           },
         },
+        {
+          title: 'Creating new scripts for Typescript in package.json',
+          enabled: (ctx: MigrateCommandContext): boolean => !ctx.skip,
+          task: async () => {
+            addPakcageJsonScripts(options.basePath, {
```
lynchbomb, line 292: typo

lynchbomb (changes requested):

packages/cli/src/commands/migrate.ts
haochuan: Ah cool thanks for catching this!

lynchbomb (approved):

PR #504, opened 2022-12-07
refactor: replace @ts-ignore with @ts-expect-error

### LINKED ISSUE #495: Consider using @ts-expect-erorr
If I'm not mistaken we should likely emit comments with `@ts-expect-erorr` instead of `@ts-ignore`. If the code doesn't compile correctly it means that we added a `@ts-expect-erorr` in a spot where it was not required which seems like a bug.

Below is an example where we are emitting a directive where it's not needed. This may help with our consumers giving us some reproductions of bugs.

With directive:

Screen Shot 2022-12-05 at 10 16 05 PM

Without directive:
Screen Shot 2022-12-05 at 10 18 19 PM

lynchbomb:
Closes #495 

This PR can optionally be merged; however I would recommend we move away from @ts-ignore. The API for @ts-ignore and @ts-expect-error are nearly identical. A key delta for @ts-expect-error, in that the compiler will throw if Rehearsal has inlined a TODO when there isn't actually a compiler error. This is super helpful for when folks are cleaning up Rehearsal TODO's and forget to delete the Rehearsal comment after fixing the error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "url": "git+https://github.com/rehearsal-js/rehearsal-js.git"
   },
   "scripts": {
+    "build": "pnpm recursive run build",
     "lint": "pnpm lint:package-json && pnpm recursive run lint",
     "lint:package-json": "sort-package-json package.json packages/*/package.json",
     "prepare": "pnpm recursive run prepare && husky install",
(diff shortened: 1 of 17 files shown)


chadhietala:
Will also help us identify where we have wrongly applied a fix which is a bug.

chadhietala (approved):

Issue #496, opened 2022-12-06
Sarif Viewer shows "[object Object]" for fixes


chadhietala:
If this is because of the propertybag, whe should think about how to render it.

<img width="821" alt="Screen Shot 2022-12-05 at 10 29 23 PM" src="https://user-images.githubusercontent.com/183799/205806255-ace3716c-f8c0-4802-a276-b2874e110646.png">

PR #513, opened 2022-12-09
fix: remove sarif property bag

brendaHuang:
[#496](https://github.com/rehearsal-js/rehearsal-js/issues/496) 

Tried using the Fixes property of sarif, but Fixes won't show in sarif reader. So I am sticking to property bag, which is more straightforward. 

Please refer to the attachments for "after" and "before" screenshots.

### DIFF
--- a/packages/reporter/src/formatters/sarif-formatter.ts
+++ b/packages/reporter/src/formatters/sarif-formatter.ts
@@ -93,7 +93,7 @@ export class SarifFormatter {
   }
 
   private buildResult(item: ReportItem): Result {
-    const { locations, relatedLocations, fixes } = this.getFilesData(
+    const { locations, relatedLocations, codeFix } = this.getFilesData(
       item.files,
       item.analysisTarget
     );
@@ -115,7 +115,7 @@ export class SarifFormatter {
       relatedLocations,
       properties: {
         fixed: item.fixed || false,
-        fixes,
+        codeFix,
       },
     };
   }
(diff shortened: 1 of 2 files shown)


brendaHuang:
Screen Shot 2022-12-08 at 6 47 51 PM

brendaHuang:
Screen Shot 2022-12-08 at 6 54 40 PM

packages/reporter/src/formatters/sarif-formatter.ts
```diff
@@ -124,30 +124,27 @@ export class SarifFormatter {
...
+          codeFixAction: file.codeFixAction || undefined,
+        };
       }
     });
 
     return {
       locations,
       relatedLocations,
-      fixes,
+      codeFix,
```
lynchbomb, line 147: want to be sure we want to change the return type to codeFix object vs an array of fixes. do we not need all the fixes anymore in this sarif entry?

lynchbomb (changes requested):

packages/reporter/src/formatters/sarif-formatter.ts
brendaHuang: I was debating this as well. If we have an array of fixes, the sarif reader won't display the fix object in the array. So this fixes or codeFixes has to be an object that's one level deep. One workaround is to append the file name to each of the 'filename', 'newCode', 'oldCode', 'codeFixAction', for example, 'fileName_xx', 'newCode_xx', 'oldCode_xx', 'codeFixAction_xx'. This way, if we have multiple files affected, we can scale to express that, though the current fix in this PR works for our current use cases. If we have more than multiple codefixes within 1 file, we can do 'filename_xx_1', 'newCode_xx_1', 'oldCode_xx_1, 'codeFixAction_xx_1'. Please let me know if this works. @lynchbomb

lynchbomb: > though the current fix in this PR works for our current use cases
ok lets keep it as implemented in this PR and expand when needed.

lynchbomb (approved):

PR #524, opened 2022-12-14
fix: codefix diagnostic ordering

### LINKED ISSUE #500: Diagnostics Ordering
I think we're going to need to control the ordering of codefixes for the diagnostics. Please see the video below. It seems we want to execute an codefix to populate parameters first and then do a right to left scan of lines fixing diagnostics.

https://user-images.githubusercontent.com/183799/205981071-fb91b4e9-694d-47ef-93b5-105aa7f3c852.mov

egorio:
Sorts diagnostics by the `start` position with prioritization of diagnostic have codes in `prioritizedCodes`.

If the diagnostic has the code mentioned in the `prioritizedCodes` list, it will be moved to the start and will be ordered against other prioritized codes in the order codes provided in the `prioritizedCodes`.

Closes #500


lynchbomb (approved):

lynchbomb:
once the failing tests are resolved this is good to merge

PR #529, opened 2022-12-15
fix: use relative paths in reports

chadhietala:
We have to use relative paths from the `basePath` so that we don't capture filesystem paths. Also systems like SonarQube wants relative paths so that we can associate issues related to a given path.


chadhietala:
I'm massively confused on how the tests pass locally but fail in CI.

PR #537, opened 2022-12-16
fix: fix format flags

### LINKED ISSUE #530: CLI `--report` should be called `--format`
We always produce a report, it's the format that changes. 

- [x] Change `--report <reportType>` flag to `--format <format>` in for migrate
- [x] Change `--report <reportType>` flag to `--format <format>` in upgrade

chadhietala:
Fixes #530


lynchbomb (changes requested):
Yeah I agree this should be `format`. You'll just need to also update the tests that are passing `-r` and `--report`

lynchbomb (approved):

PR #556, opened 2022-12-23
refactor: separate diagnostic fix from check to fix line numbers

### LINKED ISSUE #536: SARIF Report Points To Incorrect Information
We really need to make sure that the produced reports accurately point to the parts of the code that need to be modified. Currently the report is not useful because it points to random ranges in lines.

Screen Shot 2022-12-15 at 10 46 33 PM

brendaHuang:
Closes #536 

This PR does a few things:
1. Separate Diagnostic Check Plugin from the original Diagnostic Fix Plugin. The difficulty of pointing to the correct location of issues lies in the fact that we recorded the line number and column number while Diagnostic Fix Plugin was still modifying code. Running Lint Plugin against code after reporting line number may also cause issues. 
2. The revised flow goes like this: Diagnostic Fix Plugin (applying code fixes) => Lint Plugin (formatting code fixes) => Diagnostic Check Plugin (adding comments for and reporting diagnostics that cannot be fixed) => Lint Plugin(correctly indenting comments)
3. Changes to reporting: We lose information on the code fixes. We only report the result of running Diagnostic Check, which  is the remaining issues that were not fixed by Diagnostic Fix Plugin. There have been a lot of code changes on the reporting side to accommodate these changes.
4. Eslint rule change. The rule '@typescript-eslint/no-inferrable-types' is turned off to accommodate the new flow. This rule will remove the `: number` in `let count: number = 5;`, but if `let count = 5;` follows a JSDoc declaration, TypeScript will always generate a TS80004 error, with the message 'Move JSDoc declaration to type definition', which caused an existing test to fail. Please let me know if there's a better solution. 

Screenshot 2022-12-23 at 12 05 59 AM


packages/plugins/src/data/utils.ts
```diff
@@ -19,12 +11,11 @@ export function getLocation(sourceFile: SourceFile, start: number, length: numbe
...
 
-  //bump 0 to 1, so on and so forth, so that line and column can be correctly displayed in sarif.
   return {
-    startLine: startLine + 1,
-    startColumn: startColumn + 1,
-    endLine: endLine + 1,
-    endColumn: endColumn + 1,
+    startLine: startLine + INDEX_BUMP + COMMENT_BUMP,
+    startColumn: startColumn + INDEX_BUMP,
+    endLine: endLine + INDEX_BUMP + COMMENT_BUMP,
```
chadhietala, line 17: I noticed in the attached images that red squiggly seems to be off. Something is likely off in this `getLocation` function

chadhietala (changes requested):

packages/plugins/src/data/utils.ts
brendaHuang: Fixed. End column reported by Typescript was erroneously assumed to be the last affected character of the diagnostic node, however, the end column reported is the last affected character + 1. Please see new screen shot.

brendaHuang:
Screenshot 2022-12-23 at 9 32 54 AM

lynchbomb (approved):

lynchbomb:
@brendaHuang once the merge conflicts are resolved this should be good to merge

packages/cli/src/commands/migrate/tasks/convert.ts
```diff
@@ -46,14 +46,11 @@ export function convertTask(options: MigrateCommandOptions, logger: Logger): Lis
...
-          reporter.report
-        );
+        const { totalErrorCount, hintAddedCount } = getReportSummary(reporter.report);
         const migratedFileCount = migratedFiles.length;
         task.title = `${migratedFileCount} JS ${
           migratedFileCount === 1 ? 'file' : 'files'
-        } has been converted to TS. There are ${totalErrorCount} errors caught by rehearsal:
-                - ${errorFixedCount} have been fixed automatically by rehearsal.
+        } has been converted to TS. There are ${totalErrorCount} errors caught by rehearsal
                 - ${hintAddedCount} have been updated with @ts-expect-error @rehearsal TODO which need further manual check.`;
```
chadhietala, line 54:
```suggestion
        } has been converted to TS. There were ${totalErrorCount} errors caught by rehearsal:
                - ${totalErrorCount - hintAddedCount} have been fixed automatically by rehearsal.
                - ${hintAddedCount} have been updated with @ts-expect-error which need to be manually migrated.`;
```

chadhietala (changes requested):
Just a small fix. We know how many things we fixed by just subtracting the total from the hints.

packages/cli/src/commands/migrate/tasks/convert.ts
brendaHuang: @chadhietala In this PR, totalErrorCount is the total error count after we auto fixed the original errors. totalErrorCount here is in most cases identical to hintAddedCount. We are currently not reporting anything from auto fix. I can add code in another PR in diagnostic-fix plugin to report the number of original errors before auto fix. Please let me know if it's OK.

chadhietala (commented):
LGTM. Please rebase and I'll approve.

chadhietala (approved):

PR #557, opened 2022-12-23
feat: more correct formatting

chadhietala:
This PR checks to see if prettier is installed and alters `FormatCodeOptions` accordingly. Doing so will insert code from CodeActions formatted code against these adjusted options.

### DIFF
--- a/packages/codefixes/package.json
+++ b/packages/codefixes/package.json
@@ -33,6 +33,8 @@
     "@rehearsal/utils": "workspace:*"
   },
   "devDependencies": {
+    "@types/prettier": "^2.7.2",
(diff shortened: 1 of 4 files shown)


packages/codefixes/package.json
```diff
@@ -30,9 +30,11 @@
     "test:watch": "vitest --coverage --watch"
   },
   "dependencies": {
+    "@types/prettier": "^2.7.2",
```
lynchbomb, line 33: did you intentionally want to bundle this? should be a devDep

lynchbomb (changes requested):

lynchbomb (approved):

PR #573, opened 2023-01-05
fix: migrate cmd handle listr async tasks

lynchbomb:
This PR fixes bugs related to promises not being handled from Listr. We can now properly render the running tasks within the Listr stream and no longer clobber.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -51,7 +51,7 @@
   },
   "devDependencies": {
     "@types/which": "^2.0.1",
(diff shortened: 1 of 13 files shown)


haochuan (approved):
💯 💯 💯

haochuan:
@lynchbomb Just found this `default` renderer would mess up the `Edit file` option during interactive mode (see the screenshot if I select `Edit` with `$EDITOR=nvim`. 

Screenshot 2023-01-04 at 17 46 42

We can get your stuff in first and I will take a look how to fix this.

Issue #523, opened 2022-12-13
eslint report in rehearsal report


lynchbomb:
For both `upgrade` and `migrate` after eslint is configured with typescript and at the end of all CodeAction(s) we should run eslint with report output. If we can output with a SARIF formatter that would be ideal.

Issue #508, opened 2022-12-08
Explicit `any` usage should be reported


chadhietala:
Right now when the codefixes run they emit `any` which we want people to remove. Typically we would report this usage via a lint rule since TypeScript allows the explicit usage of `any` even in `strict` mode. 

We need to figure out the best way of emitting this usage into the report. Some options:

1. Custom codefix
2. Use Eslint and somehow convert the Eslint report into a sarif entry (Checkup might be a good thing to look at)

lynchbomb:
We should just NOT explicitly apply a CodeAction with `any` types. If we cannot infer the type we should not apply anything for this CodeAction.

egorio:
Currently, CodeAction can add a few types at once, so it's not very useful and easy to filter out those CodeFixes.
After the fix with `any` types applied, the next step the hint comment will be added for any.

So we can change hint with better explanation and remove @ts-ignore / @ts-expect-error to keep error to be triggered.

PR #574, opened 2023-01-05
feat: add eslint plugin

### LINKED ISSUE #568: Not All Diagnostics Should Be Reported As Errors
In #556 we moved to just report errors but we are actually marking different categories of diagnostics as errors. For example this is a suggestion diagnostic code to use the language servers refactorings to convert the function to use async await. 

Screenshot 2023-01-03 at 4 27 50 PM

I believe that [this filtering](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/plugins/src/plugins/diagnostic-check.plugin.ts#L169) needs to account for this.

brendaHuang:
Fixes #568 #508 #523 

What this PR does:
1. Add TSLintPlugin to detect typescript-related lint issues and add these issues to report, including the use of `any`. TSLintPlugin runs at the end of migrate and upgrade, so the line numbers are settled before we run the plugin.
2. Change the name of the original LintPlugin to FormattingPlugin to differentiate the purposes of these two eslint-related plugins. 
3. Filter TS diagnostics so that only those with category as 'error' will get a comment added in the file, and will be reported. 
4. Create new `.eslintrc.js` or update existing `.eslintrc.js`

Screenshot 2023-01-04 at 1 38 32 PM


packages/plugins/src/plugins/diagnostic-check.plugin.ts
```diff
@@ -71,7 +72,9 @@ export class DiagnosticCheckPlugin extends Plugin {
       }))
       .filter(
         (diagnostic) =>
-          this.isValidDiagnostic(diagnostic) && this.hasNotAddedDiagnosticComment(diagnostic, tag)
+          this.isValidDiagnostic(diagnostic) &&
+          this.isErrorDiagnostic(diagnostic) &&
```
brendaHuang, line 76: Filter out diagnostics that are not errors.

lynchbomb:
This is great! Nice job!

lynchbomb:
Rehearsal shouldn't be explicitly inlining `any` type; however on the chance it gets through its good to have a error in the report. 

We need to be sure rehearsal is leveraging the eslintrc settings from the consuming app and not rehearsal own config.

lynchbomb (approved):

packages/plugins/src/plugins/ts-lint.plugin.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    try {
+      const overrideConfig = {
+        plugins: ['@typescript-eslint'],
+        extends: ['plugin:@typescript-eslint/recommended'],
+      };
+      const options = {
+        fix: false,
+        overrideConfig,
+        useEslintrc: false,
```
lynchbomb, line 20: We def want to consume the eslintrc (comment)

lynchbomb (changes requested):

packages/plugins/src/plugins/ts-lint.plugin.ts
lynchbomb: We can also create a curated eslint config which will extend an existing https://eslint.org/docs/latest/user-guide/configuring/configuration-files#extending-configuration-files

eg. rehearsal:recommended

chadhietala:
I tried this out locally but I'm not seeing any lint violations in the report. How are you testing this?

brendaHuang:
@chadhietala Please see this screenshot for some of the eslint errors for ember-cli-pemberly-m3/addon/services/m3-schema.ts. Also in the rehearsal-js tests, additional blocks of eslint errors newly show up for existing tests. 
Screenshot 2023-01-05 at 10 32 20 AM

brendaHuang:
@chadhietala I will add a count of ts lint errors in the report to increase the visibility of ts lint errors.

packages/plugins/src/plugins/ts-lint.plugin.ts
brendaHuang: Sounds good.

brendaHuang:
@lynchbomb @chadhietala When handling lint issues, we can mimic how diagnostics are handled: first round auto fixing fixable diagnostics, second round checking and reporting remaining unfixable diagnostics. 

For lint issues, we do the same: first round auto fixing lint issues, second round checking remaining unfixable lint issues. 

The flow goes like this: 

1. **auto fixing diagnostics** (line numbers may change)
2. **auto fixing lint issues, reformatting** (line numbers may change)
3. **checking and reporting unfixable diagnostics, adding comments** (line numbers may change, but change is predictable, because we make sure the added comment is only one line, so we can report line numbers)
4. **auto fixing lint issues, reformatting, checking and reporting unfixable lint issues** (line numbers will not change, since auto fixing lint issues is mainly indenting, so we can report line numbers)

chadhietala:
I got it to work on that project but not `upload-manager`. Also it seems like the lines are not correct because we pushing the items into the report before formatting it seems.

Screenshot 2023-01-05 at 2 53 51 PM

brendaHuang:
@chadhietala added eslint config to pre-emptively address the formatting issue. 
Screenshot 2023-01-06 at 2 46 53 PM

packages/plugins/src/plugins/ts-lint.plugin.ts
brendaHuang: For now I hardcoded an essential eslint config, and add a TODO to move it to `rehearsal:recommended`. Will open a PR later to address it.

packages/cli/package.json
```diff
@@ -37,6 +37,7 @@
     "compare-versions": "^5.0.1",
     "debug": "^4.3.2",
     "enquirer": "^2.3.6",
+    "eslint": "8.22.0",
```
lynchbomb, line 40: `^8.31.0` is latest

packages/cli/src/commands/migrate/tasks/config-lint.ts
```diff
@@ -1,7 +1,10 @@
+import { resolve } from 'path';
+import { existsSync } from 'fs-extra';
+import { updateLintConfig, createLintConfig } from '../../../utils';
```
lynchbomb, line 3: lets move these functions into this file, rather than being in utils.

packages/cli/src/commands/migrate/tasks/config-lint.ts
```diff
@@ -10,7 +13,14 @@ export async function lintConfigTask(): Promise<ListrTask> {
         task.output = `Create .eslintrc.js from config`;
         await _ctx.userConfig.lintSetup();
       } else {
-        task.skip(`Skip creating .eslintrc.js since no custom config is provided`);
+        const configPath = resolve(options.basePath, '.eslintrc.js');
```
lynchbomb, line 16: the gist of these functions is the following:

1. rehearsal provides a minimal configured eslint for typescript as a default
2. [optional] if the repo isn't currently using eslint then we include the dev-dep with eslintrc from step 1
3. [optional] if the repo is already using eslint we extend the eslintrc with the rehearsal eslintrc from step 1
4. [optional] the repo in question is providing a rehearsal userConfig hook for eslint. in this case we want to run the hook AND be sure after installing the userConfig we extend with the eslintrc from step 1

packages/cli/src/utils.ts
```diff
@@ -434,3 +435,63 @@ export async function openInEditor(filePath: string): Promise<void> {
     );
   }
 }
+
+export async function updateLintConfig(path: string): Promise<void> {
```
lynchbomb, line 439: move `updateLintConfig` and `createLintConfig` into the `config-lint` file

lynchbomb (changes requested):

packages/cli/src/utils.ts
```diff
@@ -434,3 +435,63 @@ export async function openInEditor(filePath: string): Promise<void> {
...
+
+async function formatLintConfig(configStr: string, filePath: string): Promise<string | undefined> {
+  const overrideConfig = {
+    plugins: ['prettier'],
+    extends: ['plugin:prettier/recommended', 'prettier'],
+  };
+  const lintOptions = {
+    fix: true,
+    overrideConfig,
+    useEslintrc: false,
```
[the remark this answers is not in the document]
brendaHuang: Good point. I made the said change in LintCheck Plugin, but forgot to make the same change here. Will fix it.

packages/cli/src/utils.ts
brendaHuang: fixed

packages/cli/package.json
brendaHuang: fixed

packages/cli/src/utils.ts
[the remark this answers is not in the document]
brendaHuang: fixed

packages/cli/src/commands/migrate/tasks/config-lint.ts
```diff
@@ -10,8 +13,66 @@ export async function lintConfigTask(): Promise<ListrTask> {
...
+    extends: oldConfig.extends
+      ? [
+          ...oldConfig.extends,
+          'plugin:@typescript-eslint/recommended',
+          'plugin:@typescript-eslint/eslint-recommended',
+        ]
+      : ['plugin:@typescript-eslint/recommended', 'plugin:@typescript-eslint/eslint-recommended'],
+  };
+  const configStr = 'module.exports = ' + JSON.stringify(newConfig, null, 2);
+  outputFileSync(path, configStr);
```
chadhietala, line 46: I don't think we should clobber over the applications `.eslintrc.js`. It's likely better to create a different file path and we let the user know we created the file and are using it for linting and formatting.

lynchbomb: A few ways we can handle this:

1. create a new file for rehearsal-eslint in the root of the project with our curated set of plugins. add an extends in the root .eslintrc with a path to this newly created file eg.
```
  extends: [
    'plugin:@typescript-eslint/eslint-recommended',
    'plugin:@typescript-eslint/recommended',
    './.rehearsal-eslintrc.js',
  ],
```

2. read the root .eslintrc and clone it into the new rehearsal-eslint file, leaving the root .eslintrc alone. now the rehearsal-eslint

3. create a new eslint plugin eg. rehearsal and add the dependency and add the extend to the root .eslintrc

brendaHuang: Done

packages/cli/src/commands/migrate/tasks/config-lint.ts
brendaHuang: Fixed per discussion.

packages/plugins/src/plugins/ts-lint.plugin.ts
brendaHuang: fixed per discussion

chadhietala (approved):

lynchbomb (approved):

Issue #567, opened 2023-01-03
No Files Found For Files In Root


chadhietala:
Steps to reproduce:

1. mint clone upload-manager
2. `rehearsal migrate`
3. See that it finds no files

I believe the fix is that [`includePatterns`](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migration-graph-shared/src/entities/package.ts#L77) here needs to have `main` and potentially export maps to find the entry point.

wondersloth:
This has been fixed by #575

chadhietala:
I'm reopening this because the PRs still need to be merged. #595 will close this ticket.

PR #575, opened 2023-01-05
feat: include/exclude patterns; add package scope

### LINKED ISSUE #439: Bug on a JS files shared by multiple packages when migrating
Found a kinda critical issue during local test:

When `foo.js` is used by multiple packages (here "used" means that a single JS files appears in the migration file list from `migration-graph`), after the migration of the first package, `foo.js` would be converted to `foo.ts`, thus for migrations of other packages, it would throw an error that `foo.js` could not be found.

I don't have a good idea about how to deal with it at this moment though.

wondersloth:
* Adding package scope/boundary fixes #439.
  * If a file is out of the package directory is is omitted from the module graph.
* Issue #567 is a problem with the include/exclude patterns for a package. To resolve this we need to always include the package.json files properties, and include all `**/*.js` files for a plain JS library.

### DIFF
--- a/packages/migration-graph-ember/src/entities/ember-app-package-graph.ts
+++ b/packages/migration-graph-ember/src/entities/ember-app-package-graph.ts
@@ -10,6 +10,7 @@ import {
   Package,
   PackageGraph,
   PackageGraphOptions,
+  PackageJson,
   PackageNode,
 } from '@rehearsal/migration-graph-shared';
 import debug from 'debug';
@@ -25,7 +26,19 @@ class SyntheticPackage extends Package implements IPackage {
 
   constructor() {
     super('/dev/null');
+
     this.#graph = new Graph<ModuleNode>();
+
+    const EMPTY_PACKAGE_JSON: PackageJson = {};
+
+    this.pkg = new Proxy(EMPTY_PACKAGE_JSON, {
+      get(...args) {
+        const prop = args[1];
+        throw new Error(
+          `Unable to retrieve ${prop.toString()}, synthetic package created, package.json does not exist.`
+        );
+      },
+    });
   }
 
   getModuleGraph(): Graph<ModuleNode> {
@@ -34,8 +47,8 @@ class SyntheticPackage extends Package implements IPackage {
 }
 
 export type EmberAppPackageGraphOptions = {
-  parent?: GraphNode<PackageNode>;
-  project?: EmberAppProjectGraph;
+  parentNode?: GraphNode<PackageNode>;
+  projectGraph?: EmberAppProjectGraph;
   resolutions?: { services: Record<string, string> };
 } & PackageGraphOptions;
 
@@ -50,8 +63,8 @@ export class EmberAppPackageGraph extends PackageGraph {
     super(pkg, options);
 
     this.package = pkg;
-    this.parent = options.parent;
-    this.project = options?.project;
+    this.parent = options?.parentNode;
+    this.project = options?.projectGraph;
 
     this.serviceLookup = new Map<string, string>();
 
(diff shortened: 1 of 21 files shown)


packages/migration-graph-shared/src/entities/package-graph.ts
```diff
@@ -93,6 +95,13 @@ export class PackageGraph {
 
       const sourcePath = resolveRelative(baseDir, m.source);
 
+      if (this.isFileExternalToPackage(sourcePath)) {
+        DEBUG_CALLBACK(`${sourcePath} is out of package, not adding to module graph.`);
+        // Should resolve path completely relativeto the project and find which package it belongs to.
+        // Ask project which package does this belong maybe create an edge?
+        return;
+      }
+
```
wondersloth, line 104: This is the thing that resolve #439

wondersloth: If we create an edge we potentially create a cycle since the root package.json is typically a parent to all found packages within a workspace.

packages/migration-graph-ember/src/entities/ember-app-package-graph.ts
```diff
@@ -25,7 +26,19 @@ class SyntheticPackage extends Package implements IPackage {
...
+    const EMPTY_PACKAGE_JSON: PackageJson = {};
+
+    this.pkg = new Proxy(EMPTY_PACKAGE_JSON, {
+      get(...args) {
+        const prop = args[1];
+        throw new Error(
+          `Unable to retrieve ${prop.toString()}, synthetic package created, package.json does not exist.`
+        );
+      },
+    });
```
chadhietala, line 41: Is this an invariant of the system?

packages/migration-graph-shared/src/entities/package-graph.ts
chadhietala: Should we warn in these situations?

packages/migration-graph-ember/src/entities/ember-app-package.ts
```diff
@@ -12,24 +12,26 @@ export type EmberPackageOptions = PackageOptions;
 
 export class EmberAppPackage extends Package implements IPackage {
   constructor(pathToPackage: string, options: EmberPackageOptions = {}) {
-    super(pathToPackage, options);
-  }
-
-  get excludePatterns(): Array<string> {
-    // TODO Determine ember-config directory from package.json entry
-    const files = [
+    const excludePatterns = [
```
chadhietala, line 15: What happens if there are other files in the root like `.babelrc.js` or `.eslint.config.js`?

packages/migration-graph-shared/src/entities/package-graph.ts
wondersloth: I can aggregate these issues or console.warn immediately.

packages/migration-graph-ember/src/entities/ember-app-package.ts
wondersloth: We can ignore all `.` dot files.

packages/migration-graph-ember/src/entities/ember-app-package-graph.ts
wondersloth: Not sure I understand the comment. 

Like we should never try to access this on a SyntheticPackage? Then yes.

I wanted to ensure that I get a hard failure in a situation that would try to ask a SyntheticPackage for anything from it's package.json.

packages/migration-graph-ember/src/entities/ember-app-package.ts
wondersloth: By default it ignores hidden files. Adding a test to validate this.

packages/migration-graph-shared/src/entities/package-graph.ts
wondersloth: Yes we should warn, we don't have a strategy for yielding information to the user when we construct a graph. 

I think in another PR I can refactor the migration-graph packages to capture and store messages and populate that in the migration-strategy that is used in the CLI.

Alternatively a simple console.warn for now can be a temp solution.

wondersloth:
@chadhietala 

I've added a test for hidden files. We still need a strategy around config files. Running this branch of migration-graph against upload-manager I am getting this output

chadhietala (approved):

Issue #570, opened 2023-01-04
Migrate interactive mode doesn't seem to work


lynchbomb:
https://github.com/rehearsal-js/rehearsal-js/pull/558

I've tried to migrate multiple different repo's with interactive mode, however they all crash and exit with git not exiting clean. I *am* able to migrate just fine without interactive mode.

```
✖ Converting JS files to TS: Command failed with exit code 128: git diff /Users/malynch/D/dummy-ts-app/src/basic/app.js
fatal: ambiguous argument '/Users/malynch/D/dummy-ts-app/src/basic/app.js': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
error:   Error: Command failed with exit code 128: git diff /Users/malynch/D/dummy-ts-app/src/basic/app.js
fatal: ambiguous argument '/Users/malynch/D/dummy-ts-app/src/basic/app.js': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

haochuan:
Addressed in #572  

There is a bug during replace `js` file extension name with `ts`. When a file path contains `js` in the middle of the string, the code would replace that `js` instead of the actual file extension. This PR uses regex to detect and replace the actual file extension.

haochuan:
Addressed in #572

chadhietala:
@haochuan I'm still seeing this when I tried to Edit a file.

<img width="978" alt="Screenshot 2023-01-05 at 11 04 42 AM" src="https://user-images.githubusercontent.com/183799/210825786-5a87073a-8fbe-40d4-9722-c8fdc7fff1a1.png">

haochuan:
The issue mentioned by @chadhietala is caused by having default `EDITOR` with args: `EDITOR="code w"`.

This [`openInEditor`](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/utils.ts#L423-L436) needs to be changed to:
- Handle potential args in `EDITOR`
- Handle potential duplicated args for `vscode`, since currently we are appending `--wait` if the `EDITOR` is `code`

PR #577, opened 2023-01-05
fix: launch editor issue && show summary message in interactive mode

haochuan:
For the launch editor issue in the comment of #570.
Also added the summary message in interactive mode when migration completes:

Screenshot 2023-01-05 at 15 52 11


chadhietala (approved):
💥

PR #595, opened 2023-01-10
BREAKING CHANGE(migration-graph): include all files in package dir by default

### LINKED ISSUE #567: No Files Found For Files In Root

wondersloth:
This fixes #567

Currently default lookup is `index.js` only. For an repo like `upload-manager` that has loose files in the rootDir we need to look at all files.

- This PR re-applies changes from #575
- Major difference is that we are using `.` to scope the files rather than `**/*.js` and including globbing all files declared from `package.json` `files` property.
- These changes don't regress performance against `upload-manager`.

This PR does not ignore config files (webpack.config.js, eslint etc). A future PR will address this problem.

This PR depends on #596


wondersloth:
- [ ] This needs to be rebased on a different branch

packages/cli/src/helpers/report.ts
```diff
@@ -93,8 +93,9 @@ export function getReportSummary(report: Report, migratedFileCount: number): str
...
   const totalErrorCount = report.items.length;
-  return `Migration Complete. ${migratedFileCount} JS ${
-    migratedFileCount === 1 ? 'file' : 'files'
-  } has been converted to TS. There are ${totalErrorCount} errors caught by rehearsal.\n
-  ${hintAddedCount} have been updated with @ts-expect-error @rehearsal TODO which need further manual check.`;
+
+  return `Migration Complete\n\n
+  ${migratedFileCount} JS ${migratedFileCount === 1 ? 'file' : 'files'} converted to TS\n
+  ${totalErrorCount} errors caught by rehearsal\n
+  ${hintAddedCount} @ts-expect-error @rehearsal TODO which need further manual check`;
```
wondersloth, line 100: This looks like a linting change?

wondersloth:
Needs rebase.

wondersloth:
- [ ] This needs to add `main`data from package.json
- [ ] Need to explore potentially export maps as well.

egorio (approved):

wondersloth:
This branch is set to merge into root-files, I want that other PR to get landed first.

lynchbomb (approved):

PR #596, opened 2023-01-10
fix: package scope

### LINKED ISSUE #439: Bug on a JS files shared by multiple packages when migrating

wondersloth:
This PR fixes #439

In the case where a package in a workspace imports a file outside of the package scope, we will console.warn and omit the file from the package's module graph.

Potential future iteration of this would be:
- Improve the logging to be a report #603 
- Provide some interactive resolution
- Resolve the file to the actual package (typically root). This is not good because it would typically infer a cycle in the package dependency graph.


packages/migration-graph-shared/src/entities/package-graph.ts
```diff
@@ -111,9 +120,19 @@ export class PackageGraph {
           return;
         }
 
-        const relativePath = resolveRelative(baseDir, d.resolved);
+        const targetPath = resolveRelative(baseDir, d.resolved);
+        const packageName = this.package.packageName;
 
-        const dest = this.addNode({ key: relativePath, path: relativePath });
+        if (this.isFileExternalToPackage(targetPath)) {
+          console.warn(
```
lynchbomb, line 127: same as above. the user will never see this.

lynchbomb (changes requested):

packages/migration-graph-shared/src/entities/package-graph.ts
wondersloth: I'd like to address this in a later PR, this issue will address it #603

packages/migration-graph-shared/src/entities/package-graph.ts
```diff
@@ -95,6 +95,15 @@ export class PackageGraph {
 
       const sourcePath = resolveRelative(baseDir, m.source);
 
+      if (this.isFileExternalToPackage(sourcePath)) {
+        console.warn(
```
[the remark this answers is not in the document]
wondersloth: I'd like to address this in a later PR, this issue will address it #603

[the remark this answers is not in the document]
wondersloth: As far as a migration experience goes, the file will be migrated, just not by the package it was imported from.

In the **voyager** use case, where it's a big ember app with a lot of **in-repo-addon**, they have addons that import a util file from the root of the project, and use it locally in the **in-repo-addon**.

This effectively cuts off that boundary at the package scope, and **should** 🤞 be migrated by the root package once it's processed.

packages/migration-graph-shared/test/entities/project-graph.test.ts
```diff
@@ -96,6 +96,94 @@ describe('project-graph', () => {
...
+
+      // Validate graph order correctness
+      const nodes = projectGraph.graph.topSort().map((node) => node.content.pkg);
+
+      expect(nodes.length).toBe(3);
+
+      expect(nodes[0].packageName).toBe('@some-workspace/leaf');
+      expect(flatten(nodes[0].getModuleGraph().topSort())).toStrictEqual([
+        'lib/impl.js',
+        'index.js',
```
wondersloth, line 180: This should be a TODO or not included because the expected output doesn't include the `build.js` file because we currently scope all packages to traverse `index.js` only, rather than `.`. This is the test case for a workspace with a out of bounds file.



In a future PR i will be making the default entrypoint for a package `.` which will help in the use case.

packages/migration-graph-shared/test/entities/project-graph.test.ts
```diff
@@ -96,6 +96,94 @@ describe('project-graph', () => {
       expect(barNode.adjacent.has(bazNode)).toBe(true);
       expect(barNode.adjacent.has(blorpNode)).toBe(true);
     });
+    test('should not include a file out of the package scope', () => {
```
wondersloth, line 99:
```suggestion
    test.todo('should not include a file out of the package scope', () => {
```

lynchbomb (approved):

PR #600, opened 2023-01-11
refactor: remove unnecessary plugins/data folder

brendaHuang:
The original plugins/data folder was used to store the functions that collect data about code fixes. Now we don't report data about ts errors/warnings that have been fixed, so we can remove the folder.


packages/plugins/src/plugins/re-rehearse.plugin.ts
```diff
@@ -37,3 +37,21 @@ export class ReRehearsePlugin extends Plugin {
     return [fileName];
   }
 }
+
+function getBoundaryOfCommentBlock(
```
egorio, line 41: Why not be a part of the class then?

egorio (approved):

packages/plugins/src/plugins/re-rehearse.plugin.ts
brendaHuang: good point.

Issue #601, opened 2023-01-11
--dry-run option for migrate CLI


haochuan:
Adding a `--dry-run` option in `rehearsal migrate` would be useful for dev and testing things out.

wondersloth:
`rehearsal migrate --dry-run`

Should print out a list of all the files that will be attempted to migrate. 

This way a user can verify if that the target files are correct and not including directories they wish to ignore.

This will be useful when we add expose a rehearsal config to users to customize which files or directories to include/exlucde.

PR #608, opened 2023-01-12
feat: add dry run option in migrate CLI

haochuan:
For #601

Adding `-d`/`--dryRun` option in migrate CLI, which will print all the files that will be attempted to be migrated, and skip all the other tasks.

Screenshot 2023-01-12 at 11 43 11

Screenshot 2023-01-12 at 11 42 31

### DIFF
--- a/packages/cli/src/commands/migrate/index.ts
+++ b/packages/cli/src/commands/migrate/index.ts
@@ -37,6 +37,7 @@ migrateCommand
   .option('-u, --userConfig <custom json config for migrate command>', 'path to rehearsal config')
   .option('-i, --interactive', 'interactive mode')
   .option('-v, --verbose', 'print debugging logs')
+  .option('-d, --dryRun', 'print files that will be attempted to migrate', false)
   .action(async (options: MigrateCommandOptions) => {
     await migrate(options);
   });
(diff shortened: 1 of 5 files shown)


egorio:
@haochuan @wondersloth  can we use some different param name instead of `dryRun`? 
I think `dryRun` is too broad for migrate command.

haochuan:
> @haochuan @wondersloth can we use some different param name instead of `dryRun`? I think `dryRun` is too broad for migrate command... for example in upgrade it means change files but not commit and this is why I'm planning to rename it as well :)
> 
> I'm not sure about the actual name, but we can have something like :
> 
> ```
> doNotMigrate
> doNotCommint
> ```
> 
> or something like
> 
> ```
> steps=*** (filter or list)
> ```

I like the idea of `steps`, so that we can run individual tasks to check the output/result. For this case, we are just printing out the result(context) of the `initialization` task.

egorio:
> However, for this use case (check files for migration), it's a little overkill to have more generic steps=*** and I don't see too much benefits of having it at least for now. I'd prefer just having a different name.

100% agree, but have no idea what param name to use instead :)

egorio:
.... or we can keep `dryRun` but the letter include the other steps (without modifying files) - list of files to migrate, list of files to create (tsconfig, etc.), probably number of errors to fix.

haochuan:
Synced with @egorio we are going to use `--checkFiles` for now, and we are probably going to have more like this such as `--checkErrors` in the future.

chadhietala:
Chatted to @lynchbomb  about this ... `--checkFiles` seems to indicate that you're going to check the files. It's probably better to keep `--dry-run` and just list the files and then skip the rest of the steps for now. We can then iterate and flesh out the other stuff @egorio mentioned in another PR.

wondersloth:
I think it's fine to keep `--dryRun` for migrate and `--dryRun` for upgrade. It's common CLI convention to have this be a no mutative task.

For now it makes sense to just dump the files to migrate.

In future versions, it might make sense to make the `rehearsal migrate --dryRun` argument migrate files, but do not augment them in place.

That way we can surface errors or walk through an interactive mode but leave the work directory clean.

haochuan:
Cool seems like most folks agree on keep `--dryRun` as what it is for now, I've reverted my previous `--checkFiles` commit brought the origin `--dryRun` back.

chadhietala (approved):

Issue #610, opened 2023-01-12
Always install @types/node during migration


haochuan:
Need to have `@types/node` installed as a dependency for all migration use cases.

PR #611, opened 2023-01-12
feat: install @types/node

haochuan:
For #610


packages/cli/src/commands/migrate/tasks/dependency-install.ts
```diff
@@ -3,6 +3,17 @@ import { ListrTask } from 'listr2';
 import { addDep } from '../../../utils';
 import type { MigrateCommandContext, MigrateCommandOptions } from '../../../types';
 
+export const REQUIRED_DEPENDENCIES = [
```
egorio, line 6: <nerd>I would sort them, but it's fine :)</nerd>

egorio (approved):

packages/cli/src/commands/migrate/tasks/dependency-install.ts
haochuan: I like it, sorted 😄

PR #612, opened 2023-01-13
refactor: migrate as pull based gen and itr

lynchbomb:
This PR is a refactor of migrate and implements pull based plugins and file iteration via generators/iterators. 

The initial goal was this would mitigate the SIGINT issue with the prev blocking implementation, however doesn't appear to resolve it.


chadhietala (approved):
Pull based FTW

lynchbomb:
Screenshot 2023-01-13 at 12 08 31 PM

Seeing what's up with this. This test is failing in CI, but not failing locally.

PR #617, opened 2023-01-14
chore: remove `src` from tarballs

egorio:
We don't need `src` folder to be published, [package-json#files](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files)


chadhietala (approved):
Good cleanup!

PR #618, opened 2023-01-14
chore: refactor to use Package.path

wondersloth:
There were three different methods for accessing Package.path. This consolidates everything to the `Package.path` getter.

### DIFF
--- a/packages/migration-graph-shared/src/entities/package.ts
+++ b/packages/migration-graph-shared/src/entities/package.ts
@@ -21,7 +21,7 @@ export type PackageContainer = {
 };
 
 export type PackageOptions = {
-  type?: string;
+  packageType?: string;
   packageContainer?: PackageContainer;
   rootPackagePath?: string;
   name?: string;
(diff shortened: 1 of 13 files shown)


packages/migration-graph-shared/src/entities/package.ts
```diff
@@ -267,14 +252,14 @@ export class Package implements IPackage {
...
-    const pathToPackageJson = join(this.path, 'package.json');
+    const pathToPackageJson = join(this.#packagePath, 'package.json');
     writeJsonSync(pathToPackageJson, sorted, { spaces: 2 });
   }
 
   isConvertedToTypescript(conversionLevel?: string): boolean {
     const fastGlobConfig = {
       absolute: true,
-      cwd: this.path,
+      cwd: this.#packagePath,
```
wondersloth, line 262: This should just be this.path for simplicty.

lynchbomb (approved):
Good refactor!

PR #619, opened 2023-01-14
chore(migration-graph-ember): tidy types

wondersloth:
- Creates a type `EmberProjectPackage` which is `EmberAppPackage | EmberAddonPackage | Package`.
- Helps reduce misuse around the code base.
- Consolidate types to a single file.


chadhietala (approved):
Seems like a good cleanup.

Issue #602, opened 2023-01-11
Create rehearsal config for migrate


wondersloth:
**Problem**
There is no way for a **user** (migrator) to customize what files get include/excluded in the migrate experience.

For a plain js library, there are various setups and conventions we have to work around:
- `build` vs `dist` output directory.
- What file we should not convert like configs (e.g.  `eslint.config.*`, `webpack.config.*`, `karma.config`)
- Include/exclude directories (e.g. `test/tests`, `docs`, `docusaurus`, `ember-backstop`);

We have some naive defaults, but we should provide a config API for customizing the input.

**Proposal**

Provide a config to the user will allow for the user (migrator) to tune the output of the migraiton-graph to meet their needs.

We to pass this config from the `cli` package to `migration-graph`, so that when a package is crawled we can include/exclude specific directories and files.

Currently we have a naive set of include/exclude filters defined in the constructors of a Package:
- `Package` https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migration-graph-shared/src/entities/package.ts#L82
- `EmberAppPackage` https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migration-graph-ember/src/entities/ember-app-package.ts#L17
-  `EmberAddonPackage` https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migration-graph-ember/src/entities/ember-addon-package.ts#L25

- [x] A way to configure each package or just for the project, which files or directories to ignore.
  - Configure by package, all packages, or at the project level.
- [x] Leverage existing config work:
  - https://linkedin.ghe.com/multiproduct/rehearsal-cli-config
  - https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/user-config.ts
- [x] Improve rehearsal-config.json to have a schema
  - This can be a defined schema similar to [tracerbench tb-schema.json](https://raw.githubusercontent.com/TracerBench/tracerbench/master/packages/cli/tb-schema.json)
  - Using this package https://www.npmjs.com/package/typescript-json-schema
- [x] If types are shared we should move this all to `util` package.

PR #626, opened 2023-01-17
feat(migration-graph): ignore more files

wondersloth:
When regressing migration graph against our internal list of JS libaries, these were the common files that should be ignored from the graph when we forgo passing an `--entrypoint` argument.

The following will be ignored by default

**Directories**
```
.yarn // yarn3 directory
dist
test
tests
```

```
.eslintrc.*
.babelrc.*
babel.config.* // Babel configs
Brocfile.js 
prettier.config.* // Prettier configs
karma.config.*
webpack.config.js
vite.config.ts
```

This is ultimately a short term fix until this lands: https://github.com/rehearsal-js/rehearsal-js/issues/602

### DIFF
--- a/packages/migration-graph-shared/src/entities/package.ts
+++ b/packages/migration-graph-shared/src/entities/package.ts
@@ -79,7 +79,25 @@ export class Package implements IPackage {
       this.#packageContainer = { isWorkspace: () => false };
     }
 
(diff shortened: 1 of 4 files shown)


lynchbomb:
Yeah this is totally ok for now. Eventually we might want to expose the ability to override this hardcoded list via rehearsal config.

lynchbomb (approved):

PR #628, opened 2023-01-18
feat(migration-graph): refactor console.warn statements to debug

wondersloth:
- Expose `MigrationStrategy.report` that contains the logged messages from the migration-graph creation process.
- This implementation passes through and down logger references from the top level `migration-graph` package.


packages/migration-graph-ember/test/utils/discover-ember-service-dependencies.test.ts
```diff
@@ -29,6 +30,28 @@ describe('discoverServiceDependencies', () => {
     expect(results).toBeFalsy;
   });
 
+  test('should optionally log an error if  afile cannot be parsed', () => {
```
wondersloth, line 33:
```suggestion
  test('should optionally log an error if a file cannot be parsed', () => {
```

packages/migration-graph-shared/src/utils/logger.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.#logs.push({ severity: 'log', message });
+  }
+}
+
+// Module scoped logger;
+const logger: Logger = new Logger();
+
+export function getLogger(): Logger {
+  return logger;
+}
```
wondersloth, line 35: We are not using this it can be removed.
```suggestion
```

wondersloth:
Should I be using Winston logger here or is that overkill?

lynchbomb:
> Should I be using Winston logger here or is that overkill?

@wondersloth what specifically are you thinking about for the migration-graph report and logger? eg. what would it look like to the user?

wrt logger: _if_ we were to propagate something from the migration-graph into Listr you would need access to the ListrContext eg. [like we do here](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migrate/src/migrate.ts#L248-L251). If we just want to have access to the logs for debug purposes I would just use the `DEBUG` module. 

wrt report: you would need access to the reporter eg. [like we do here](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migrate/src/migrate.ts#L60)

You shouldn't need Winston logger at all.

wondersloth:
> > Should I be using Winston logger here or is that overkill?
> 
> @wondersloth what specifically are you thinking about for the migration-graph report and logger? eg. what would it look like to the user?
> 
> wrt logger: _if_ we were to propagate something from the migration-graph into Listr you would need access to the ListrContext eg. [like we do here](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migrate/src/migrate.ts#L248-L251). If we just want to have access to the logs for debug purposes I would just use the `DEBUG` module.
> 
> wrt report: you would need access to the reporter eg. [like we do here](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migrate/src/migrate.ts#L60)
> 
> You shouldn't need Winston logger at all.

Ok, so I think I am going to convert my existing console.log or logger statements in this PR to just a `DEBUG` module call. 

There is some value in maybe adding some interactivity to the ember service discovery code, but I don't think that needs to be addressed.

I need to 
- [ ] Remove logger from this PR.
- [ ] Replace logger calls with `DEBUG_CALLBACK` calls.
- [ ] Rewrite tests to spy on `DEBUG` to validate those cases.

packages/migration-graph/src/migration-strategy.ts
```diff
@@ -36,13 +35,28 @@ export class SourceFile {
 }
 
 export class MigrationStrategy {
-  #files: Array<SourceFile> = [];
+  #rootDir: string;
   #sourceType: SourceType;
+  #files: Array<SourceFile> = [];
+  #messages: Array<string> = [];
```
wondersloth, line 41: Remove this

packages/migration-graph-ember/src/utils/discover-ember-service-dependencies.ts
```diff
@@ -46,8 +46,8 @@ export function discoverServiceDependencies(
       decoratorsBeforeExport: true,
     });
   } catch (e) {
-    console.warn(`Ember service discovery failed. Unable to parse: ${filePath}`);
-    DEBUG_CALLBACK(`Error: %0`, e);
+    // Push into project messages;
+    DEBUG_CALLBACK(`Ember service discovery failed. Unable to parse: %s\n%s`, filePath);
```
wondersloth, line 50:
```suggestion
    DEBUG_CALLBACK(`Ember service discovery failed. Unable to parse: %s`, filePath);
```

packages/migration-graph/src/migration-strategy.ts
wondersloth, line 41:
```suggestion
```

packages/migration-graph/src/migration-strategy.ts
```diff
@@ -36,13 +35,28 @@ export class SourceFile {
...
+    return this.#rootDir;
+  }
+
+  get report(): Array<string> {
+    return this.#messages;
+  }
+
+  set report(messages: Array<string>) {
+    this.#messages = messages;
+  }
```
wondersloth, line 58:
```suggestion
```

packages/migration-graph/test/migration-strategy.test.ts
```diff
@@ -38,6 +38,35 @@ describe('migration-strategy', () => {
...
+
+      const expected = [
+        `[warn] The source file "build.js" is importing a file "../../some-shared-util.js" that is external to "@some-workspace/leaf" package directory (${someDir}/packages/leaf), omitting target file ("../../some-shared-util.js") form package-graph.`,
+        `[warn] The target file "../../some-shared-util.js" is external to package "@some-workspace/leaf" (${someDir}/packages/leaf), omitting target file form package-graph.`,
+        `[warn] The source file "build.js" is importing a file "../../some-shared-util.js" that is external to "@some-workspace/branch" package directory (${someDir}/packages/branch), omitting target file ("../../some-shared-util.js") form package-graph.`,
+        `[warn] The target file "../../some-shared-util.js" is external to package "@some-workspace/branch" (${someDir}/packages/branch), omitting target file form package-graph.`,
+      ];
+
+      expect(actual).toStrictEqual(expected);
+    });
```
wondersloth, line 69:
```suggestion
```

packages/migration-graph-ember/src/utils/discover-ember-service-dependencies.ts
```diff
@@ -46,8 +46,8 @@ export function discoverServiceDependencies(
       decoratorsBeforeExport: true,
     });
   } catch (e) {
-    console.warn(`Ember service discovery failed. Unable to parse: ${filePath}`);
-    DEBUG_CALLBACK(`Error: %0`, e);
+    // Push into project messages;
```
wondersloth, line 49:
```suggestion
```

lynchbomb (approved):

Issue #644, opened 2023-01-19
Migrate CLI doesn't respect yarn/npm version with volta


haochuan:
During rehearsal migrate initialization task, if a machine uses `volta` to control the package manager's version, [getModuleManagerInstaller](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/utils.ts#L237) would always user the default global version, instead of using the one specified in the project.

For example, if `yarn` in `package.json` has been pinned to version 3:
```
 "volta": {
  34   │     "node": "16.15.0",
  35   │     "yarn": "3.2.1"
  36   │   }
```

Inside the project:
```
❯ which yarn
~/.volta/bin/yarn

❯ yarn -v
3.2.1
```

However, in [getModuleManagerInstaller](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/utils.ts#L237), `const bin = which.sync(manager);` would be the global one `~/.volta/tools/image/yarn/1.22.10/bin/yarn` instead of `~/.volta/bin/yarn`. This would cause a lot of potential issues for the mismatched package manager's version.

egorio:
Seems related and will be fixed as well https://github.com/rehearsal-js/rehearsal-js/issues/429

haochuan:
In any place

PR #646, opened 2023-01-19
fix: package manager version conflict with volta

haochuan:
For #644.

If a project is using `npm` or `yarn` with `volta`, use `volta which npm/yarn` to get the path of the manager bin path, instead of `which npm/yarn`.

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -191,6 +191,28 @@ export function isPnpmManager(): boolean {
   return !!pnpmPath;
 }
 
+// Check if a binary exists and if it's executable
+export function isBinExisted(binName: string): boolean {
+  try {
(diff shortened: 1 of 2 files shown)


egorio (approved):

packages/cli/src/utils.ts
```diff
@@ -191,6 +191,28 @@ export function isPnpmManager(): boolean {
...
+  try {
+    which.sync(binName);
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
+
+// Get the binary path for a package manager
+export function getManagerBinPath(
```
egorio, line 205: I used this way to check: You can do async like this: https://github.com/rehearsal-js/action/blob/master/src/pm.ts (check --version) :)

PR #650, opened 2023-01-21
fix: release script

egorio:
Improve release.js script

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "rehearsal-workspace",
+  "version": "1.0.2-beta",
   "private": true,
   "repository": {
     "type": "git",
--- a/release.js
+++ b/release.js
@@ -61,15 +61,15 @@ function publish() {
   console.log(gitClean);
   execSync(gitClean);
 
-  // replace the version in package json on all packages with version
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "rehearsal-workspace",
+  "version": "1.0.2-beta",
```
egorio, line 3: This is needed for `pnpm changelog`

release.js
```diff
@@ -61,16 +61,16 @@ function publish() {
...
 
-  // replace the version in package json on all packages with version
-  const bumpVersion = `pnpm -r version ${newVersion} --exact --no-git-tag-version`;
-  console.log(bumpVersion);
-  execSync(bumpVersion);
-
-  // pnpm install recursive
-  const pnpmInstall = `pnpm -r install`;
+  // pnpm install
+  const pnpmInstall = `pnpm install`;
```
egorio, line 65: Found that we don't need to install recursively

release.js
```diff
@@ -61,16 +61,16 @@ function publish() {
...
   console.log(pnpmInstall);
   execSync(pnpmInstall);
 
+  // pnpm prepare
+  const pnpmPrepare = `pnpm prepare`;
```
egorio, line 70: This restores husky

Issue #593, opened 2023-01-10
"fix" CLI flag to specify which curated set of diagnostics to run


lynchbomb:
To achieve this we need to https://github.com/rehearsal-js/rehearsal-js/issues/314

Right now for the CLI migrate command we are running transforms against the kitchen sink of diagnostics after a curated set of [priority order diagnostics](https://github.com/rehearsal-js/rehearsal-js/blob/1d4170567611bbc78746bd59affe0e4200ccde2a/packages/plugins/src/plugins/diagnostic-fix.plugin.ts#L18-L29) 

In order for us to determine the value add of running ONLY the curated set vs the curated + kitchen sink we should expose a CLI flag eg. `rehearsal migrate --infer=minimal`.

Not sure specifically what we should call this; more so we need something like this.

egorio:
We can do #591 and then we can implement `--infer=minimal` to compare sets of fixes.

egorio:
@lynchbomb I'm going to do "minimal" (safe to applu codefixes) by default and `--infer=all` will be used instead.

PR #653, opened 2023-01-22
feat: limit a list of typescript codefixes safe to apply

egorio:
Limit the amount of TypeScript-provided fixes we use based on `fixName`
Filter out text changes containing `:any` types.


packages/codefixes/src/base-codefix-collection.ts
```diff
@@ -11,11 +11,17 @@ export class BaseCodeFixCollection implements CodeFixCollection {
     this.list = list;
   }
 
-  getFixForDiagnostic(diagnostic: DiagnosticWithContext): CodeFixAction | undefined {
-    if (this.list[diagnostic.code] !== undefined) {
-      return this.list[diagnostic.code].getCodeAction(diagnostic);
+  getFixesForDiagnostic(diagnostic: DiagnosticWithContext): CodeFixAction[] {
```
egorio, line 14: Providers to return all available code fixes (not only the first one)

packages/codefixes/src/codefixes-provider.ts
```diff
@@ -12,17 +12,14 @@ export class CodeFixesProvider {
...
-  getCodeFixes(diagnostic: DiagnosticWithContext): CodeFixAction | undefined {
+  getCodeFixes(diagnostic: DiagnosticWithContext, onlySafeFixes: boolean): CodeFixAction[] {
+    let fixActions: CodeFixAction[] = [];
     for (const collection of this.collections) {
-      const action = collection.getFixForDiagnostic(diagnostic);
-      // Return the first available CodeFixAction
-      if (action !== undefined) {
-        return action;
-      }
+      fixActions = [...fixActions, ...collection.getFixesForDiagnostic(diagnostic, onlySafeFixes)];
```
egorio, line 20: Merge all available codefixes from all providers to display to user in the interactive mode

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
```diff
@@ -16,16 +16,19 @@ export class DiagnosticFixPlugin extends Plugin {
...
-    7006, // Parameter implicitly has an ___ type
-    7008, // Member implicitly has an ___ type
-    80004, // JSDoc types may be moved to TypeScript types
-    90016, // Declare property
-    90035, // Declare private property
-    90053, // Declare a private field named
-    2525, // Initializer provides no value for this binding element and the binding element has no default value
+    80002, // convertFunctionToEs6Class
+    80005, // requireInTs
+    80004, // annotateWithTypeFromJSDoc
```
egorio, line 21: Prioritized `types form jsDocs` over `from usage` because it seems easy to fix incorrect types in jsDocs or remove them to use types `from usage` instead of relying on the compiler.

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
```diff
@@ -43,25 +46,30 @@ export class DiagnosticFixPlugin extends Plugin {
...
-          let text = this.service.getFileText(fileTextChange.fileName);
+      }
 
-          const textChanges = normalizeTextChanges([...fileTextChange.textChanges]);
-          for (const textChange of textChanges) {
-            text = applyTextChange(text, textChange);
-          }
+      // Use the first available codefix in automatic mode,
+      // TODO: User should be able to choose one of the fixes form this list in interactive mode
+      const fix = fixes.shift()!;
```
egorio, line 57: @lynchbomb we will be able to interact with users in this place to allow them to choose which fix to apply :)

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
```diff
@@ -43,26 +46,33 @@ export class DiagnosticFixPlugin extends Plugin {
         continue;
       }
 
-      const fix = codefixes.getCodeFixes(diagnostic);
-      if (!fix) {
+      const fixes = codefixes.getCodeFixes(diagnostic, true, false);
```
chadhietala, line 49: The [countdown latch](https://github.com/rehearsal-js/rehearsal-js/pull/653/files#diff-11e3e65b5b4fefbcc62fbc0826e0036b9a8755b79ad29a33ff8149b126a17152R41) is going to be triggered for all `Diagnostic`s that are of the kind `Suggestion` when you filter out [codefixes](https://github.com/rehearsal-js/rehearsal-js/pull/653/files#diff-e3a9a4e5319af2f6b16a9a5b453cb236d0be0ce253205801aacb8851a6ea9953R110)

Please see this example https://tsplay.dev/Wylpnm

I believe you're going to need to unify getting the diagnostics and the codefixes.

packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -16,7 +16,52 @@ import type { CodeFixCollection, DiagnosticWithContext } from './types';
 export class TypescriptCodeFixCollection implements CodeFixCollection {
   hasPrettier: boolean | undefined;
   prettierConfigs: PrettierOptions | undefined;
-  getFixForDiagnostic(diagnostic: DiagnosticWithContext): CodeFixAction | undefined {
+
+  // List of safe to apply code fix names
+  // @see ./helpers/available-fix-names.ts for the list of available fixNames
```
lynchbomb, line 21: similar comment - after we move `available-fix-names.ts` out of src to docs or even README.md link it here

packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -16,7 +16,52 @@ import type { CodeFixCollection, DiagnosticWithContext } from './types';
...
+  safeCodeFixNames = [
```
lynchbomb, line 22: We should move this to a separate file so we can eventually import a given set of `safeCodeFixNames` based on the CLI argument (for future API)

lynchbomb: eg https://github.com/rehearsal-js/rehearsal-js/issues/593

lynchbomb (changes requested):

packages/codefixes/src/typescript-codefix-collection.ts
egorio: P. S. I keep this list in the provider because different providers have different fixNames. I tried to keep it in the plugin, but it become maintenance more complicated.

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
lynchbomb: ^^^^ @egorio in your example. What happens when you also have jsDoc comments?

egorio: Non-any types will be applied, `any` skipped and the comment added. 
Added test cases to cover `any`

packages/codefixes/src/typescript-codefix-collection.ts
egorio: Moved to .md

packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -47,11 +92,54 @@ export class TypescriptCodeFixCollection implements CodeFixCollection {
...
+
+    for (const changes of fix.changes) {
+      for (const textChanges of changes.textChanges) {
+        // Don't return dummy function declarations
+        if (textChanges.newText.includes('throw new Error')) {
+          return false;
+        }
+
+        // Don't return `: any` types as part of text changes
+        const anyTypeRegex = /[a-zA-Z0-9_$?]: any/;
```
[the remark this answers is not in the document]
egorio: Removing any is not a trivial task :) 
Removed this behavior from the current PR, will be added in the future.

packages/codefixes/src/types.ts
```diff
@@ -29,8 +29,16 @@ export interface CodeHint {
   hint: string;
 }
 
+export interface CodeFixCollectionFilter {
```
egorio, line 32: We have codefixes files now. Can be extended in the future for any user experience improvement purpose :)

packages/plugins/src/plugins/diagnostic-check.plugin.ts
```diff
@@ -23,9 +23,8 @@ export class DiagnosticCheckPlugin extends Plugin {
     DEBUG_CALLBACK(`Plugin 'DiagnosticCheck' run on %O:`, fileName);
 
     const allFixedFiles: Set<string> = new Set();
-    let tries = diagnostics.length + 1;
```
egorio, line 26: Tries were needed for cases when @ts-ignore didn't work.
We don't rely on it anymore and don't need this countdown anymore.

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
```diff
@@ -16,53 +16,67 @@ export class DiagnosticFixPlugin extends Plugin {
...
+    7044, // inferFromUsage
+    7045, // inferFromUsage
+    7046, // inferFromUsage
+    7050, // inferFromUsage
+    2612, // addMissingDeclareProperty
   ];
 
   async run(fileName: string): PluginResult {
     let diagnostics = this.getDiagnostics(fileName);
-    let tries = diagnostics.length + 1;
```
egorio, line 33: This was added for cases where @ts-ignore didn't work at the time this plugin added comments.
This counter doesn't work anymore (especially with prioritized diagnostics) and the only possible case it is needed is (hypothetically)  when we continuously have the same error after applying the same fix (if we have a fix)... but we will work on the issue once we will face this case.

packages/upgrade/test/fixtures/upgrade/ts-any.ts.output
```diff
@@ -1,8 +1,56 @@
+/**
```
egorio, line 1: Some of `any` cases.

packages/upgrade/test/fixtures/upgrade/ts-any.ts.output
```diff
@@ -1,8 +1,56 @@
...
+  return b;
+}
+
 function capitalize(str: string): string {
   return str[0].toUpperCase() + str.substring(1);
 }
 
-export function doSomething(args: string | any[], name: string) {
+/* @ts-expect-error @rehearsal TODO TS7006: Parameter 'args' implicitly has an 'any' type. */
+export function doSomething(args, name: string) {
```
egorio, line 53: `args` type is detected as `string | any[]`, but we drop the whole suggested fix because it contains any.
We can do some work around this in the next PRs if needed.

lynchbomb (approved):

PR #661, opened 2023-01-25
feat: add --regen option

### LINKED ISSUE #438: feat: rehearsal migrate --report
This feature is to track the state of where a project is in a migration. There are 3 scenarios we need to solve for:

### Scenario 1: Tightening Over Time

In this scenario a team has ran `rehearsal migrate` across their entire code base and have a SARIF report file in their code base. As teams start to fix issues a couple of things will happen:

1. `@ts-expect-error` comments will be removed as they are fixed
2. Fixing issues and editing the project's file will shift the line and col numbers around causing the report to point to the wrong thing

### Scenario 2: Partial Migration

In this scenario an application team has chosen not to move to TypeScript all in one approach but instead are taking a more piecemeal approach. This is a superset of scenario 1 but with the added edition of needing to know how many more files need to be converted to TS.

## Tasks

Because making progress and tightening types up over time causes issues with the report and people may be taking different approaches to migration we need the ability to re-generate the report so it is accurate. By adding this functionality we also open up the ability for application teams to use cron or pre-commit hooks to ensure the report is fresh.

I believe the tasks are something like the following although this may not be fully exhaustive:

- [ ] Gather up all the TypeScript & JavaScript files if there are any files in the project
- [ ] Make migrate aware of the `--report` flag
- [ ] Use the [ReRehearsePlugin](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/plugins/src/plugins/re-rehearse.plugin.ts) to remove all the `@ts-expect-error @rehearsal TODO` comments
- [ ] Run the FormatPlugin on the TS files
- [ ] Update the [diagnostic-check.plugin.ts](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/plugins/src/plugins/diagnostic-check.plugin.ts) or create a new plugin that pushes entries into the report  to track any unmigrated JS files. These probably need to be reported or tagged differently.
- [ ] Run the [diagnostic-check.plugin.ts](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/plugins/src/plugins/diagnostic-check.plugin.ts) to collect all the existing errors
- [ ] Write the report back

## Done

To close out this feature we need to be able to regenerate the report to track the progress of a migration by both fixing `@ts-expect-error` comments and files migrated

## Open Questions

1. By adding JS file tracking into the mix we diverging as to what `rehearsal migrate` gives you by the default. We need to think about how to resolve that.

brendaHuang:
Closes #438 

1. regenerate migration report, with or without uncommitted code changes
2. check if ts config and eslint config exist before generating report
3. scan ts files to check for errors
4. what regen does: 
  - remove existing rehearsal comments
  - check ts errors and insert comments
  - reformatting
  - check lint errors
  - sending ts errors and lint errors to reporter

Screenshot 2023-01-24 at 8 58 55 PM

### DIFF
--- a/packages/cli/src/commands/migrate/index.ts
+++ b/packages/cli/src/commands/migrate/index.ts
@@ -1,10 +1,16 @@
 #!/usr/bin/env node
+import { resolve } from 'path';
 import { Command } from 'commander';
 import { Listr } from 'listr2';
 import { createLogger, format, transports } from 'winston';
 
 import { version } from '../../../package.json';
-import { parseCommaSeparatedList, gitIsRepoDirty, resetFiles } from '../../utils';
+import {
+  parseCommaSeparatedList,
+  gitIsRepoDirty,
+  resetFiles,
+  getLintConfigPath,
+} from '../../utils';
 import {
   initTask,
   depInstallTask,
@@ -12,6 +18,7 @@ import {
   tsConfigTask,
   lintConfigTask,
   createScriptsTask,
+  regenTask,
 } from './tasks';
 
 import type { MigrateCommandOptions } from '../../types';
@@ -38,6 +45,7 @@ migrateCommand
   .option('-i, --interactive', 'interactive mode')
   .option('-v, --verbose', 'print debugging logs')
   .option('-d, --dryRun', 'print files that will be attempted to migrate', false)
+  .option('-r, --regen', 'print out current migration status')
   .action(async (options: MigrateCommandOptions) => {
     await migrate(options);
   });
@@ -50,7 +58,7 @@ async function migrate(options: MigrateCommandOptions): Promise<void> {
 
   console.log(`@rehearsal/migrate ${version.trim()}`);
 
-  if (!options.dryRun) {
+  if (!options.dryRun && !options.regen) {
     const hasUncommittedFiles = await gitIsRepoDirty(options.basePath);
     if (hasUncommittedFiles) {
       logger.warn(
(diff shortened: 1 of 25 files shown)


packages/cli/src/commands/migrate/index.ts
```diff
@@ -38,6 +45,7 @@ migrateCommand
   .option('-i, --interactive', 'interactive mode')
   .option('-v, --verbose', 'print debugging logs')
   .option('-d, --dryRun', 'print files that will be attempted to migrate', false)
+  .option('-r, --regen', 'print out current migration status')
```
wondersloth, line 48: Thinking of naming on this.

Does it make more sense to name this `--rerun`.

This use case (to my understanding) is that I've run `rehearsal migrate` and now I am wanting to regen a report of errors?

What is the by-product of this argu `--regen` does it only create a new sarif report?

Just thinking through this, it's more of a `--status` and it updates the report?

wondersloth (commented):
Overall looks good to me. I defer to @lynchbomb on final ship as he probably has more context on the feature.

wondersloth:
Some linting errors in you CI/ Runs
```
packages/regen lint: error TS18003: No inputs were found in config file '/home/runner/work/rehearsal-js/rehearsal-js/packages/regen/test/tsconfig.json'. Specified 'include' paths were '["**/tmp/**/*"]' and 'exclude' paths were '[]'.
```

packages/cli/src/commands/migrate/index.ts
brendaHuang: @wondersloth In my current implementation, the difference between `migrate --regen` and `migrate` is that `migrate` also includes autofixes, and `migrate --regen` does not. `migrate --regen` will only mark the ts errors with comments, and print out ts errors and eslint errors in report. But I do see your point. What you do think, @lynchbomb ?

brendaHuang:
@wondersloth Config file lint error fixed.

packages/cli/src/commands/migrate/index.ts
lynchbomb: (Lets keep `regen`. This isn't technically a rerun since it doesn't actually migrate/upgrade anything. We only want to reproduce the report.

lynchbomb (approved):

PR #666, opened 2023-01-30
feat: attempt to download types

chadhietala:
This PR implements the pieces so that we can attempt to download types of the diagnostic code 7016


packages/plugins/src/plugins/diagnostic-fix.plugin.ts
```diff
@@ -59,20 +59,30 @@ export class DiagnosticFixPlugin extends Plugin {
       // TODO: User should be able to choose one of the fixes form this list in interactive mode
       const fix = fixes.shift()!;
 
-      for (const fileTextChange of fix.changes) {
-        let text = this.service.getFileText(fileTextChange.fileName);
+      if (isInstallPackageCommand(fix)) {
+        try {
+          await this.service.getLanguageService().applyCodeActionCommand(fix.commands[0]);
```
egorio, line 64: Why only `[0]`?

I though about another loop before `for (const fileTextChange of fix.changes) {`

```
for (const command of fix.commands) {
   await this.service.getLanguageService().applyCodeActionCommand(command);
}

for (const fileTextChange of fix.changes) { 
  ...
}
```

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
```diff
@@ -59,20 +59,30 @@ export class DiagnosticFixPlugin extends Plugin {
...
+        } catch (e) {
+          // swallow the error we couldn't install the dep or it didn't actually exist
+        }
+      } else {
```
egorio, line 68: I think we need to run `fix.changes` after commands are applied because after the package is installed it can modify files with `import` statements.

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
```diff
@@ -143,3 +153,9 @@ export class DiagnosticFixPlugin extends Plugin {
     });
   }
 }
+
+function isInstallPackageCommand(
```
egorio, line 157: We probably don't need this, because of currently CodeActionComand == InstallPackageAction :)

image

but can keep it for future stability :) 

but can keep

egorio: Also, I think

`isInstallPackageCommand(command: CodeActionCommand)`
instead 
`isInstallPackageCommand(fix: CodeFixAction)`

seems better :)

packages/migrate/test/fixtures/migrate/src/complex.js
```diff
@@ -1,4 +1,7 @@
 import { salutations, DEFAULT_GREETING } from './salutations';
+import { v4 as uuid } from 'uuid';
+
+uuid();
```
brendaHuang, line 4: Need to add Line 2-4 to packages/migrate/test/fixtures/migrate/output/complex.ts.ouput.

egorio (approved):

egorio:
Thank you!

PR #672, opened 2023-01-30
fix: fetch eslint config error

### LINKED ISSUE #663: migrate CLI error out for eslint plugins when running migrate with -p option
Existed `.eslintrc.js`:

```
module.exports = {
  ...
  plugins: ['ember'],
  ...
};
```

Run `rehearsal migrate`, we have:

```
› <project-path>/.eslintrc.js already exists, extending Rehearsal default typescript-related config
✖ Update eslintrc.js: Failed to load plugin 'ember' declared in '.eslintrc.js': Package subpath './lib/util/traverser' is not defined by "exports" in <project-path>/node_modules/eslint/package.json
Referenced from: <project-path>/.eslintrc.js
error:   Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Failed to load plugin 'ember' declared in '.eslintrc.js': Package subpath './lib/util/traverser' is not defined by "exports" in <project-path>/node_modules/eslint/package.json
Referenced from: <project-path>/.eslintrc.js
```

brendaHuang:
Closes #663

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -37,6 +37,7 @@
     "chalk": "^4.1.2",
     "commander": "^9.4.0",
     "compare-versions": "^5.0.1",
+    "cosmiconfig": "^8.0.0",
     "debug": "^4.3.2",
     "enquirer": "^2.3.6",
     "eslint": "^8.31.0",
--- a/packages/cli/src/commands/migrate/tasks/config-lint.ts
+++ b/packages/cli/src/commands/migrate/tasks/config-lint.ts
@@ -1,8 +1,9 @@
 import { resolve } from 'path';
 import { ESLint } from 'eslint';
 import { outputFileSync } from 'fs-extra';
+import { cosmiconfigSync } from 'cosmiconfig';
 import defaultConfig from '../../../configs/default-eslint';
-import { gitAddIfInRepo, getLintConfigPath } from '../../../utils';
+import { gitAddIfInRepo, getEsLintConfigPath, determineProjectName } from '../../../utils';
 import type { ListrTask } from 'listr2';
 import type { MigrateCommandContext, MigrateCommandOptions } from '../../../types';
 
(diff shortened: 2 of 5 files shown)


packages/cli/src/commands/migrate/tasks/config-lint.ts
```diff
@@ -60,12 +64,25 @@ async function extendsRehearsalInCurrentConfig(
...
   basePath: string = process.cwd()
 ): Promise<void> {
-  /* eslint-disable-next-line @typescript-eslint/no-var-requires */
-  const oldConfig = require(configPath);
-  const newConfig = {
-    ...oldConfig,
-    extends: Array.from(new Set([...oldConfig.extends, rehearsalConfigRelativePath])),
-  };
+  const projectName = determineProjectName(basePath);
+  const explorerSync = cosmiconfigSync(projectName || '');
```
lynchbomb, line 68: Why do we need cosmicconfig? What was the bug this is resolving?

lynchbomb (changes requested):

lynchbomb (approved):

packages/cli/src/commands/migrate/tasks/config-lint.ts
brendaHuang: Originally we use `require' to retrieve the existing `eslintrc.js` in the target project. This works if we run `migrate` in the target project, but when we run `migrate` in a different directory and pass in `-p`, `require` does not work.

packages/cli/package.json
```diff
@@ -37,6 +37,7 @@
     "chalk": "^4.1.2",
     "commander": "^9.4.0",
     "compare-versions": "^5.0.1",
+    "cosmiconfig": "^8.0.0",
```
wondersloth, line 40: ~How is this dependency being used? Was it not added by accident?~
Saw @lynchbomb 's resolved comment.

wondersloth (approved):

PR #675, opened 2023-01-30
feat: plugin options

egorio:
Add plugin options to allow plugins.

This PR is not the final solution, plugins will be wrapped to PluginRunner next PR to make it more nice and clear.


packages/migrate/src/migrate.ts
```diff
@@ -1,17 +1,16 @@
...
-  DiagnosticCheckPlugin,
-  LintCheckPlugin,
-} from '@rehearsal/plugins';
+import { Plugin, RehearsalService } from '@rehearsal/service';
+import { DiagnosticCheckPlugin, DiagnosticFixPlugin, LintPlugin } from '@rehearsal/plugins';
 import { findConfigFile, parseJsonConfigFileContent, readConfigFile, sys } from 'typescript';
-import type { Reporter } from '@rehearsal/reporter';
-import type { Logger } from 'winston';
+import { PluginOptions } from '@rehearsal/service/dist/src/plugin';
+import { LintPluginOption } from '@rehearsal/plugins/dist/src/plugins/lint.plugin';
```
brendaHuang, line 8: importing from dist folder?

packages/migrate/src/migrate.ts
```diff
@@ -1,17 +1,16 @@
...
+import { DiagnosticPluginOptions } from '@rehearsal/plugins/dist/src/plugins/diagnostic-fix.plugin';
```
brendaHuang, line 9: same as above

packages/migrate/src/migrate.ts
```diff
@@ -1,17 +1,16 @@
...
+import { DiagnosticCheckPluginOptions } from '@rehearsal/plugins/dist/src/plugins/diagnostic-check.plugin';
```
brendaHuang, line 10: dist?

packages/service/src/plugin.ts
```diff
@@ -3,20 +3,16 @@ import type { Logger } from 'winston';
...
+  run(fileName: string, options: PluginOptions): PluginResult;
+}
 
-  constructor(service: RehearsalService, reporter: Reporter, logger?: Logger) {
-    this.service = service;
-    this.logger = logger;
-    this.reporter = reporter;
-  }
+export interface PluginOptions {
+  [key: string]: unknown;
```
brendaHuang, line 12: Can we add basePath to PluginOptions?

egorio: You can add this option to specific plugin, like for Lint here:
https://github.com/rehearsal-js/rehearsal-js/pull/675/files#diff-3be91568bd3d2f0e95e214fa7b01784fa345decb2da97e9211aaecc92b30bda7R7-R9

but I will think about sending PluginContext (cli, basePath, service, etc) together with PluginOptions next PR.

brendaHuang (approved):

Issue #637, opened 2023-01-19
migrate-state.json packageMap absolute path


lynchbomb:
Currently the partial migration state is handled with `migrate-state.json` with the key `packageMap` pointing to the [absolute path](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/migrate/tasks/initialize.ts#L61) of the app in question. I think we are going to have an issue when apps are doing a partial migration with handoff. The workflow:

- dev-a runs a partial migration against "project-foo" 
- They check-in the `migrate-state.json` into a eg. migrate-to-typescript branch
- dev-b picks up the partial migration work against "project-foo" from the same branch
- The `migrate-state.json` contains the absolute path to "project-foo" from dev-a eg. 
```
{
  "name": "project-foo",
  "packageMap": {
    "/Users/dev-a/project-foo": []
  },
  "files": {}
}
```

Instead we want the relative path to the project in the `packageMap`

PR #677, opened 2023-01-30
fix: use relative path in state file

haochuan:
For #637.

- Make sure all the file/package paths are relative in `migrate-state.json`
- Small cleanups
- New test in `initialize` to check the package selection menu and the state file


egorio (approved):

lynchbomb:
Fixes #637

PR #685, opened 2023-01-31
fix: provide cwd for lint plugin

### LINKED ISSUE #674: Need to fetch eslint config from target project when running migrate with -p
Currently we use eslint config from current directory to run lint plugins. This is erroneous when we run `migrate` with `-p`.

brendaHuang:
Closes #674


egorio (approved):
Great!

PR #695, opened 2023-02-03
fix: use a queue

chadhietala:
This adds a queueing mechanism to the diagnostics that ensures that we do not try to fix issues over and over again. This occurs when you have diagnostics with a kind of `Suggestion` and it is either not applied or partially applied. Examples include:

- Suggestion to download `@types/<packageName>` where `@types/<packageName>` does not exist
- Suggestion to consume JSDoc comments that will emit `any`

There is probably a better factoring around this that includes tracking what has and has not been processed when we call `getDiagnostics`, but in general this code needs to be adapted to be pull based anyways. This starts us down that road by pulling items off of a queue.


egorio:
Chad, I found this solution overcomplicated :(
Can you check this one: https://github.com/rehearsal-js/rehearsal-js/pull/698... I think should work as well

PR #698, opened 2023-02-03
chore: skip fixes that didn't work

egorio:
A simple solution for https://github.com/rehearsal-js/rehearsal-js/pull/695:

Save all processed `diagnostic + fix` and ignore them if we try to fix the same diagnostic with the same fix. 

Closes #695

### DIFF
--- a/packages/plugins/src/plugins/diagnostic-fix.plugin.ts
+++ b/packages/plugins/src/plugins/diagnostic-fix.plugin.ts
@@ -1,5 +1,6 @@
-import { CodeActionCommand, type DiagnosticWithLocation } from 'typescript';
+import { CodeActionCommand, CodeFixAction, type DiagnosticWithLocation } from 'typescript';
 import { debug } from 'debug';
+import hash from 'object-hash';
 
 import {
   codefixes,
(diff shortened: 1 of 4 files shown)


egorio:
Seems like having a few errors triggered a the same place are legit behavior :(

Here we have the same place triggers diagnostic twice, and different fixes applied each time (it another file):

```
upgrade/test/fixtures/transforms/4082-2.tsx
  start: 502, code: 4082
[ { newText: 'export ', span: { start: 140, length: 0 } } ]
upgrade/test/fixtures/transforms/4082-2.tsx
  start: 502, code: 4082
[ { newText: 'export ', span: { start: 98, length: 0 } } ]
```

Modified the code to skip processed `diagnostic + fix`.

packages/plugins/src/plugins/diagnostic-fix.plugin.ts
```diff
@@ -37,6 +38,8 @@ export class DiagnosticFixPlugin implements Plugin<DiagnosticFixPluginOptions> {
     2612, // addMissingDeclareProperty
   ];
 
+  attemptedToFix: string[] = [];
```
chadhietala, line 41: We should reset this array on either line 90 or line 44 so we don't retain the hashes for entirety of the execution. This could lead to memory issues in really large applications if we don't

egorio: Thanks. Done.

Issue #630, opened 2023-01-18
Refactor blocking code from all plugins


lynchbomb:
Similar to what we are doing in [migrate.ts](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migrate/src/migrate.ts#L124-L214) we should refactor the blocking code in all plugins eg. [here](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/plugins/src/plugins/diagnostic-check.plugin.ts#L28), [here](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/plugins/src/plugins/diagnostic-fix.plugin.ts#L38) etc. as pull based with a generator/iterator implementation. 

This will likely be critical for the state management of the migrate command. Currently with the migrate command if you control-c from interactive mode all state is lost as we dont currently yield to saving the migrated files to disk. 

In the proposed refactor we could do just this; yield -> save to disk

- [ ] refactor and remove the `while` loops from all plugins

egorio:
We have to use while loop in `-fix` plugins because each fix has side-effect on the file and program - fixing one issue can fix many others in this file or can bring new errors. We need to check the latest issues every time we modify file. 

We can think about removing the loop from `-check` (report) plugin where we just add comments (we can just add comments from the bottom top)... but will think.

I don't think while loop is causing the huge slowdown, but something inside (like [this](https://github.com/rehearsal-js/rehearsal-js/issues/890)) does it.

PR #699, opened 2023-02-04
fix: refactor blocking code from all plugins

egorio:
Use `migrate` code as an example of a non-blocking solution for `PluginsRunner` :)
Now all 3 - `migrate`, `upgrade`, and `regen` utilizes the same code... and seem can be merged together because they are basically the same (except for a number of plugins and their configs. 

Also separated PluginOptions and PluginsRunnerContext

### DIFF
--- a/packages/migrate/src/migrate.ts
+++ b/packages/migrate/src/migrate.ts
@@ -1,15 +1,8 @@
 import { existsSync } from 'fs';
 import { dirname, extname, resolve } from 'path';
 import { execSync } from 'child_process';
-import { Plugin, PluginOptions, RehearsalService } from '@rehearsal/service';
-import {
-  DiagnosticCheckPlugin,
-  DiagnosticCheckPluginOptions,
-  DiagnosticFixPlugin,
-  DiagnosticFixPluginOptions,
-  LintPlugin,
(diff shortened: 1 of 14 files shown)


egorio:
Seems like having few errors triggered a the same place are legit behavior :(

Here we have the same place triggers diagnostic twice, and different fixes applied each time (it the another file):

```
upgrade/test/fixtures/transforms/4082-2.tsx
  start: 502, code: 4082
[ { newText: 'export ', span: { start: 140, length: 0 } } ]
upgrade/test/fixtures/transforms/4082-2.tsx
  start: 502, code: 4082
[ { newText: 'export ', span: { start: 98, length: 0 } } ]
```

lynchbomb:
This is awesome! Great work. This pushes us closer to wrap up #630 - however we still need to refactor the blocking code in the plugins eg [here](https://github.com/rehearsal-js/rehearsal-js/blob/dc2c6add5b6946b1730feff66c53eea71e6db557/packages/plugins/src/plugins/diagnostic-check.plugin.ts#L43)

We can merge this - lets just not close issue #630 yet

lynchbomb (approved):

PR #700, opened 2023-02-04
feat: ignore fixes with `object` type, same was as `any`

### LINKED ISSUE #696: Ignore `Object` types same way we ignore `any`
It doesn't make sense to add `: object` type when migrating projects.

egorio:
Fixes #696


wondersloth (commented):
Can you add a test that exercises the "Object" inference for a return type. 

```
TS2339: Property 'type' does not exist on type 'object'.
```

lynchbomb (approved):

PR #702, opened 2023-02-04
feat: process yml, json eslint config

### LINKED ISSUE #688: Process existing eslint config in yaml, yml, json formats
Read, parse, and write existing eslint config in target project which is in yml, yaml, json formats in addition to js

brendaHuang:
Closes #688 

This PR handles when target project has existing `.eslintrc.yml`, `.eslintrc.yaml`, or `.eslintrc.json`. Previously we only handle `.eslintrc.js`.


chadhietala:
Can we make sure to cover `.eslintrc` which is also used. It should be the same as `.eslintrc.json` from what I've seen

packages/cli/src/commands/migrate/tasks/config-lint.ts
```diff
@@ -1,14 +1,25 @@
...
 
-const REHEARSAL_CONFIG_FILENAME = '.rehearsal-eslintrc.js';
-const REHEARSAL_CONFIG_RELATIVE_PATH = `./${REHEARSAL_CONFIG_FILENAME}`;
+enum REHEARSAL_CONFIG_FILENAMES {
+  JS = '.rehearsal-eslintrc.js',
+  YAML = '.rehearsal-eslintrc.yml',
+  JSON = '.rehearsal-eslintrc.json',
+}
+
+enum FORMAT {
```
lynchbomb, line 17: We just need to handle the legacy (deprecated but we will likely still find it) `.eslintrc` (note this is not js/yaml/json extension)

lynchbomb (changes requested):

packages/cli/src/commands/migrate/tasks/config-lint.ts
brendaHuang: @lynchbomb Just to clarify, we need to handle .eslintrc, .eslintrc.js, .eslintrc.json, .eslintrc.yaml, .eslintrc.yml. Right?

brendaHuang:
@chadhietala @lynchbomb `.eslintrc` added

packages/cli/src/commands/migrate/tasks/config-lint.ts
lynchbomb: Correct. `.eslintrc` is older - but we should handle it

lynchbomb (approved):

Issue #708, opened 2023-02-06
New versions of test-support is not published to npmjs


egorio:
[Two latest versions](https://www.npmjs.com/package/@rehearsal/test-support) of test-support packages was not published to npmjs (it is still 1.0.4-beta).

This happening because `pnpm publish` is stuck while trying to publish it.
I was able to publish it manually after removing `node_modules` from `fixtures`.

The short solution is to remove those folders before running `pnpm publish` (in release.js), but we can investigate the root cause and fix it properly.

PR #723, opened 2023-02-10
fix: release script

egorio:
The `pnpm -r publish` is stuck because of node_modules in fixtures... 
We don't need those folders to do the release e anyway.

Fixes: #708

### DIFF
--- a/release.js
+++ b/release.js
@@ -122,9 +122,9 @@ function publish() {
   execSync(gitPushTags);
 
   // publish
-  // const pnpmPublish = `pnpm -r publish`;
-  // console.log(pnpmPublish);
-  // execSync(pnpmPublish);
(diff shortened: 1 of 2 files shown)


release.js
```diff
@@ -121,10 +121,15 @@ function publish() {
   console.log(gitPushTags);
   execSync(gitPushTags);
 
+  // remove fixture's node_nodules to avoid stuck on the next step
+  const cleanFixturesNodeModules = `git clean -dfx packages/test-support/fixtures`;
```
wondersloth, line 125: This works for now. We can also just exclude this with `files` in package.json

wondersloth (approved):

release.js
egorio: Updated to solve it this way :)

PR #729, opened 2023-02-10
fix: push only new tag during release

egorio:
Small improvements


wondersloth (changes requested):
Lookgs good. Snapshots need to be updated in cli. I had the same issue with the eslint bump in my PR.

lynchbomb (approved):

PR #730, opened 2023-02-10
fix: upgrade cli tests

### LINKED ISSUE #728: `test/commands/upgrade.test.ts` augments `cli/package.json` often moving typescript from devDependency to dependency
In **packages/cli/** running:

```
$ pnpm vitest test/commands/upgrade.test.ts
```

```
diff --git a/packages/cli/package.json b/packages/cli/package.json
index 7b7b286..11d751d 100644
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -48,13 +48,13 @@
     "listr2": "^5.0.7",
     "simple-git": "^3.16.0",
     "tmp": "^0.2.1",
+    "typescript": "^4.9.5",
     "which": "^3.0.0",
     "winston": "^3.8.2",
     "yaml": "^2.2.1"
   },
   "devDependencies": {
     "@types/which": "^2.0.1",
-    "typescript": "^4.9.5",
     "typescript-json-schema": "^0.55.0",
     "vitest": "^0.28.4"
   },
(END)
```

This test should create a project using **fixturify-project** so that mutation of a package.json can happen in isolation.

egorio:
Fixes #728

### DIFF
--- a/packages/cli/test/commands/migrate/__snapshots__/e2e.test.ts.snap
+++ b/packages/cli/test/commands/migrate/__snapshots__/e2e.test.ts.snap
@@ -318,7 +318,7 @@ exports[`migrate: e2e > init flag 3`] = `
     "@types/node": "^18.13.0",
     "@typescript-eslint/eslint-plugin": "^5.51.0",
     "@typescript-eslint/parser": "^5.51.0",
-    "eslint": "^8.33.0",
+    "eslint": "^8.34.0",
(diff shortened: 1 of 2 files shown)


wondersloth (approved):
Oh, so the reason it would augment it, because the command would attempt to install against our package.json and it would remove it from devDep and it in dep. So this mirrors the package.json in CLI now.

PR #732, opened 2023-02-13
feat: reporting for sequential runs

### LINKED ISSUE #693: Sequential Runs Overwrite existing SARIF files
If you target only certain directories of a project for migration e.g.`rehearsal migrate -e foo` then `rehearsal migrate -e bar` you will overwrite the migration report. I think there are several paths forward that need to be discussed.

- Ask if we should overwrite the file
- Append into the file
- Write another file one already exists

brendaHuang:
Closes #693 

New behavior:

When we detect there are past runs in the existing report of the base path directory, we automatically run `--regen` for past runs, which are determined by basePath + entrypoint, before we run migrate itself with current entrypoint. The regenerated reports for past runs are merged into the current report. 

Changes to @rehearsal/reporter:

1. `Reporter.report` now is the aggregate of data from all runs.
2. `Reporter.currentRun` is the data from the current run. 
3. API of Reporter has been changed, such as, from `save()` to `saveCurrentRunToReport()`, `saveReport()`, etc.
4. Report summary has been changed to array to record runs of migration.
5. basePath + entrypoint saved in summary of the report for each run.
6. fixedItemCount is a cumulative sum of all errors autofixed by rehearsal in all runs.
7. json report will be automatically produced in all cases, since it has the complete raw data.
8. As a bonus, the code changes help fix an existing bug in interactive migrate. As we interactively migrate every file, we print a report for each file migrated. Since we passed in one Reporter instance for the entire migrate command, the last report generated for the last file interactively migrated has the cumulative errors (un-deduplicated). Now with the new reporter, the repeat errors are filtered, since we merge current run with existing data. That's the reason why some of the test results are changed. 

Changes to @rehearsal/migrate
1. Added Rerehearsal plugin at the start of @rehearsal/migrate to remove possible rehearsal comments added by preceding runs of `--regen`.

Screenshot 2023-02-13 at 8 13 55 AM
Screenshot 2023-02-13 at 8 26 21 AM

### DIFF
--- a/packages/cli/src/commands/migrate/index.ts
+++ b/packages/cli/src/commands/migrate/index.ts
@@ -2,6 +2,8 @@
 import { Command } from 'commander';
 import { Listr } from 'listr2';
 import { createLogger, format, transports } from 'winston';
+import { resolve } from 'path';
+import { existsSync } from 'fs';
 
 import {
   parseCommaSeparatedList,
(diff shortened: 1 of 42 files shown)


brendaHuang:
Screenshot 2023-02-13 at 1 15 34 PM

lynchbomb (approved):

PR #740, opened 2023-02-14
feat(migration-graph): improving --entrypoint cli

### LINKED ISSUE #655: Tracking 'test(s)' directory migration
We currently can migrate vanilla JS code to TS. We need to expand this functionality to include Ember usecases for both Classic and Embroider V2.

wondersloth:
Fixes #655 with tests for plain JS single package, workspaces, and ember.

### Changes
- Added `test/sample.test.js` to plain JS test directories.
- Removed `test` and `test` from default exclude directories for `Package`, `EmberAppPackage`, `EmberAddonPackage`.
- Include all dirs by default `.` for all packages (only excluding tmp, dist, node_modules etc.)
- Removed default includePattern for PackageGraph of `index.js` since it's alway `.` unless an entrypoint is passed.
- Updated all test output to have new test file in output file list.
- The `discoverEmberPackages` API exposed from `migration-graph-ember` through `migration-graph` now runs off of `EmberAppProjectGraph.discover()` and returns a flattened array of packages.

### Removed
- The following has been removed `module-mappings.ts` code and test support. Implementation into `EmberAppProjectGraph.ts`
  - Also removed `PackageContainer`, `EmberPackageContainer`
  - This also hollowed out the `EmberAppPackage` and `EmberAddonPackage` tests.
- `EmberAddonPackage` most get/set methods for addons paths are removed becaus they're no longer used.

### TODOs
- [x] Create edge between test file and impl in an ember app. Ensure resolution of `app/components/foo.js` works from a test file.


packages/migration-graph-ember/src/entities/ember-app-project-graph.ts
```diff
@@ -132,7 +79,11 @@ export class EmberAppProjectGraph extends ProjectGraph {
       deps = deps.concat(...somePackages);
     }
 
-    if (pkg instanceof EmberAppPackage) {
+    // When ever we come across an EmberAppPackage or EmberAddonPackage
+    // We look at the packageJson.ember-addon.paths and resolve those paths for the
+    // We should have all those packages already populated in our mappingsByLocation
```
wondersloth, line 84:
```suggestion
    // We look at the packageJson.ember-addon.paths and resolve those paths to
    //  the in-repp addons.
    // We should have all those packages already populated in our mappingsByLocation
```

wondersloth:
Will follow up with a PR that wires up interactive mode to the project-graph correctly.

packages/migration-graph-ember/src/entities/ember-app-project-graph.ts
```diff
@@ -132,7 +79,11 @@ export class EmberAppProjectGraph extends ProjectGraph {
...
+    // via from discovery().
+    if (isApp(pkg.packageJson) || isAddon(pkg.packageJson)) {
```
wondersloth, line 86:
```suggestion
    // When ever we come across an EmberAppPackage or EmberAddonPackage
    // We check packageJson.ember-addon.paths for additional dependencies.
    // This gives us additional dependency data not found from explicit dependency
    // and devDependency entries in package.json.
    if (isApp(pkg.packageJson) || isAddon(pkg.packageJson)) {
```

lynchbomb (approved):

PR #743, opened 2023-02-14
feat: infer return types based on inlay hints and inheritance 

### LINKED ISSUE #543: Infer return types based on inlay type hints
TS provides functions to get [Inlay Hints](https://code.visualstudio.com/docs/typescript/typescript-editing).

Those hints include:
- FunctionParameterTypeHints
- includeInlayVariableTypeHints
- includeInlayPropertyDeclarationTypeHints
- includeInlayFunctionLikeReturnTypeHints
- includeInlayEnumMemberValueHints

We can utilize them to add types where TS didn't provide CodeFixes.

For example, the next code doesn't trigger suggestions for many places where inline has types:

image

Ref:
- LanguageService.[provideInlayHints](https://github.com/microsoft/TypeScript/blob/3716ffe748fe799b1c69090056397bda814c0d49/src/services/types.ts#L596)()

Sister to: https://github.com/rehearsal-js/rehearsal-js/issues/526

egorio:
Use 7050 error to trigger 2 code fixes to infer function return types. 

Closes #543


lynchbomb:
@egorio will you provide more content here as to what specifically doesn't work as expected?

packages/codefixes/src/codefixes.ts
```diff
@@ -6,13 +6,18 @@ import { AddErrorTypeGuardCodeFix } from './fixes/addErrorTypeGuard/index.js';
 import { AddMissingExportCodeFix } from './fixes/addMissingExport/index.js';
 import { MakeMemberOptionalCodeFix } from './fixes/makeMemberOptional/index.js';
 import { AddMissingTypesBasedOnInheritanceCodeFix } from './fixes/addMissingTypesBasedOnInheritance/index.js';
+import { AddMissingTypesBasedOnInlayHintsCodeFix } from './fixes/addMissingTypesBasedOnInlayHints/index.js';
 
 export const codefixes = new CodeFixesProvider([
+  new BaseCodeFixCollection({
+    7050: new AddMissingTypesBasedOnInheritanceCodeFix(),
```
egorio, line 13: Need to be called before `7050: new AddMissingTypesBasedOnInlayHintsCodeFix(),` but can't place it to the following array because of duplicate keys. 

Thinking about refactor collection to  accept:

```
7050: [
  new AddMissingTypesBasedOnInheritanceCodeFix(),
  new AddMissingTypesBasedOnInlayHintsCodeFix(),
]
```

and few other ways... but don't want to do it in this PR.

packages/codefixes/src/fixes/addMissingTypesBasedOnInheritance/index.ts
```diff
@@ -1,32 +1,105 @@
...
+    if (!diagnostic.node) {
+      return undefined;
+    }
+
+    if (isParameter(diagnostic.node)) {
+      return this.fixMissingParameterType(diagnostic);
+    }
+
+    if (isMethodDeclaration(diagnostic.node)) {
+      return this.fixMissingReturnType(diagnostic);
```
egorio, line 22: This CodeFix now supports return types of a method declaration.

egorio:
>  will you provide more content here as to what specifically doesn't work as expected

Function param types were not great using inlay hints. The return types are opposite - looks great in provided inlay hints :)

packages/cli/package.json
```diff
@@ -68,7 +68,6 @@
     "@types/js-yaml": "^4.0.5",
     "@types/which": "^2.0.1",
     "fixturify": "^3.0.0",
-    "typescript": "^4.9.5",
```
lynchbomb, line 71: why are you removing typescript from devDeps here?

lynchbomb (changes requested):
Can you add a test scenario for an async function where the return type will include the inferred promise as `:Promise<T>`

packages/cli/package.json
egorio: One of tests does this. Already reverted.

egorio:
@lynchbomb test for an async function added

PR #748, opened 2023-02-16
fix(cli): config test; remove snapshot, assert contains

wondersloth:
### Problem
* cli test for `e2e.test.ts` were failing because the snapshots had the version of the rehearsal/cli.
* snapshots had the old version vs current version.

### Fix
This removes the snapshots form the user-config include/exclude tests, and just asserts the output contains specific data.


packages/cli/test/commands/migrate/__snapshots__/e2e.test.ts.snap
```diff
@@ -350,7 +350,7 @@ exports[`migrate: e2e > regen result after the first pass 1`] = `
 `;
 
 exports[`migrate: e2e > user defined options passed by --user-config -u > migrate.exclude 1`] = `
-"[32minfo[39m:    @rehearsal/migrate 1.0.8-beta
+"[32minfo[39m:    @rehearsal/migrate 1.0.9-beta
```
chadhietala, line 353: I believe that we need to remove the version from the output

wondersloth: I changed the test. I assert on the stdout. Will regen the snapshot and it will be gone.

lynchbomb (approved):

Issue #746, opened 2023-02-15
Smoke Testing


lynchbomb:
We need a way to smoke test the bundled rehearsal CLI in a separate JS repo. This needs to run via action against every PR and against the latest published build.

PR #758, opened 2023-02-17
test: smoke test gha

lynchbomb:
Fixes: #746 

The intention behind this PR is on every push/pr run this action: https://github.com/rehearsal-js/smoke-test/blob/master/.github/workflows/run-smoke-test.yml. 

The smoke-test-action will:

- download the master branch zip from https://github.com/rehearsal-js/rehearsal-js/archive/refs/heads/master.zip
- pack the tarball for the @rehearsal/cli from zip rather than from NPM and as file:dep
- run test suite from https://github.com/rehearsal-js/smoke-test/tree/master/test
- if the test suite passes this action should pass

The downsides with this implementation:
- we are downloading from master rather than from the PR/commit/branch. as such we will only have an integration pulse after PR merge into master (we still catch way before release and execute manually if needed)
- I would like to also implement a life cycle catch to run the smoke-test on `prepublish`

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -1,0 +1,12 @@
+name: Smoke-Test
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+  uses:
+    rehearsal-js/smoke-test/.github/workflows/run-smoke-test.yml@run-smoke-test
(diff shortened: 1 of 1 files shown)


lynchbomb:
We might need to leverage a different GHA event: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch. TBD

chadhietala (approved):

PR #763, opened 2023-02-17
feat: types based on inheritance

egorio:
When function param has missing type:
- trying to confirm that function is part of a class
- check all extended or inherited classes and interfaces to find the function with the same name 
- check if the function has a parameter with the same ~name~ index and type

Types `any` and `object` are ignored by default.

I'm pretty sure there are edge cases I didn't cover...


packages/codefixes/src/fixes/addMissingTypesBasedOnInheritance/index.ts
```diff
@@ -0,0 +1,60 @@
...
+          continue;
+        }
+
+        const resolvedType = checker.getTypeOfSymbolAtLocation(
+          ancestorMethod,
+          heritageClauseTypeNode
+        );
+
+        // Expect this to return type, but `any` is returner
+        // checker.typeToString(checker.getTypeOfSymbolAtLocation(ancestorMethod.getDeclarations()[0].parameters[0].symbol.valueDeclaration))
```
egorio, line 44: Expect to get type here, but having `any`. 

The `ancestorMethod` is part of `Animal`

packages/codefixes/src/codefixes.ts
```diff
@@ -14,4 +15,7 @@ export const codefixes = new CodeFixesProvider([
     4082: new AddMissingExportCodeFix(),
   }),
   new TypescriptCodeFixCollection(),
+  new BaseCodeFixCollection({
+    7006: new AddMissingTypesBasedOnInheritanceCodeFix(),
+  }),
```
chadhietala, line 20: Is there a reason to add this collection here? Why wouldn't we have it in the first collection?

packages/codefixes/src/fixes/addMissingTypesBasedOnInheritance/fixtures/types-based-on-inheritance.ts.input
```diff
@@ -0,0 +1,63 @@
+class Animal {
```
chadhietala, line 1: Can you add cases for:


Multiple Implements
```
interface Thing {
  thing(num: number): void
}

class MultipleImplements implements Wild, Thing {
  wild(place) { return 'foo' };
  thing(num) { console.log(num); };
}
```

Generic Interfaces
```
interface GenericThing<T = number> {
  thing(num: T): void
}

type Thing = { foo: number };

interface ConstrainedThing<T extends Thing> {
  thing(num: T): void
}

class GenericImpl implements GenericThing {
  thing(num) { console.log(num); };
}

class GenericImplTwo implements GenericThing<number> {
  thing(num) { console.log(num); };
}

class GenericImplTwo implements ConstrainedThing<{ foo: number, bar: string }> {
  thing(num) { console.log(num); };
}
```

Generic Classes
```
class GenericThing<T = number> {
  thing(num: T): void
}

type Thing = { foo: number };

class ConstrainedThing<T extends Thing> {
  thing(num: T) {
  
  }
}

class GenericImpl extends GenericThing {
  thing(num) { console.log(num); };
}

class GenericImplTwo extends GenericThing<number> {
  thing(num) { console.log(num); };
}

class GenericImplTwo extends ConstrainedThing<{ foo: number, bar: string }> {
  thing(num) { console.log(num); };
}
```

chadhietala:
💥

packages/codefixes/src/codefixes.ts
egorio: Codefixes are applied in the order defined there.
I want to try this one after TypeScrupt ones were tried to apply.

packages/codefixes/src/fixes/addMissingTypesBasedOnInheritance/fixtures/types-based-on-inheritance.ts.input
egorio: Test cases added and handled :)

lynchbomb:
Awesome work!

lynchbomb (approved):

Issue #606, opened 2023-01-11
`.gjs` support in migration graph


chadhietala:
Ember applications support a `.gjs` file extension. These files are effectively JS files with the exception in that they allow `<template>...</template>` tags within the file for declaring the template for a component. Components must but imported into the file explicitly and are called in `<template/>`s.

[Example](https://github.com/ember-template-imports/ember-template-imports/blob/master/tests/dummy/app/components/gjs-test.gjs)

These files need to be crawled and stitched into the module graph so that we can convert then from `.gjs` to `.gts` files with types.

We should be be able to use [`@glimmer/babel-preset`](https://github.com/glimmerjs/glimmer.js/tree/master/packages/%40glimmer/babel-preset) to parse the file and then start following the import declarations. Or we could try just using a [lexer](https://github.com/guybedford/es-module-lexer) that assumes the input is valid to get get the imports.

wondersloth:
@chadhietala 

> We should be be able to use [@glimmer/babel-preset](https://github.com/glimmerjs/glimmer.js/tree/master/packages/%40glimmer/babel-preset) to parse the file and then start following the import declarations. Or we could try just using a [lexer](https://github.com/guybedford/es-module-lexer) that assumes the input is valid to get get the imports.

If given a source like:
```
import Component from '@glimmer/component';

const divide = () => 4 / 2;

const First = <template>Hello</template>

class Second extends Component {
  <template>world</template>
}

<template>
  <First/>, <Second/>!
</template>
```

From what I can tell `@glimmer/babel-preset` only does manipulation on `hbs()` calls to produce the correct transformations.

We need to use https://github.com/ember-template-imports/ember-template-imports to be able to parse a source file above.

PR #772, opened 2023-02-22
feat: support .gjs in Ember PackageGraphs

wondersloth:
This PR addresses #606 

Adds support for `.gjs` files to Ember PackageGraphs.

- Includes `.gjs` file extension for `EmberAppPackageGraph` and `EmberAddonPackageGraph` resolveOptions passed to dependency-cruiser.
- Added test for `EmberAddonPackage` due to bug found where resolve alias for ember-addon paths starting with moduleName where not mapping back to the addon directory when regressing against internal addon.
- Preprocess/transform `.gjs` files to remove `<template>` tags and use something more es parse-able.

### Testing Done
- Added tests
- Regressed against internal addon `voyager-web/packages/addons/image-view-model` which has `.gjs` files. Produces graph edges correctly. 

### TODO
- [x] Add service discovery support for .gjs files.
- [x] Drop `@glimmer/babel-preset` as we don't need to transform **`hbs`**. 
- [X] Figure out how to infer `@glimmer/syntax` dependency from target package's node_modules so we can use the version related to the ember app.

### ~Concerns/Alternatives~ 
**This is no longer a concern because I removed usage of `futureCruise` api that was experimental/unstable.**

- The `futureCruise` method from `dependency-cruiser` API is experimental/unstable.
- Not currently a problem, but a potential future risk

#### Drop futureCruise API, no need for `@glimmer/babel-preset`
- This may sound odd, but we don't need `@glimmer/babel-preset` to add `.gjs` files to the graph using `dependency-cruiser`
- I was able to implement this without using `@glimmer/babel-preset` because ember  `<template>` tag semantics get seen as JSX TemplateExpressions.
- When `dependency-cruiser` parses these files, it thinks they're JSX like source files and manages to get the import declarations perfectly fine.
- First couple of commits to this PR have just that. ~Last commit of this PR is the refactor to use `futureCruise`.~ We don't use futureCruiser at all. Just plain old cruise. 

#### Find another graph solution
- Some other OTS graph creation tool.
- Pros:
  - Maybe more stable API?
- Cons:
  - Have to find and adapt another solution.
  - Uncertain if it will provide same file level data.
  - Uncertain if it will meet or exceed current perf.

#### Create our own dependency graph solution
- Expand on the naive `swc` es6 entrypoint walker I had before.
- Pros: We control our own crawl tools. Maybe faster.
- Cons:
  - Have to re-invent the wheel.
  - This is a solved problem, by mostly all package managers.
 

#### Explore and emit the resolved graph data from rollup when passed an entrypoint

Implementation:
- fast glob all entrypoints in a project `.` directories and top level files.
- Then iterate with rollup

Pros:
- If it works, a way to customize resolutions for different module environments.
- Well rationalized API.

Cons:
- Perf is uncertain.
- Uncertain of success
- Have to create a new solution.

### DIFF
--- a/packages/migration-graph-ember/package.json
+++ b/packages/migration-graph-ember/package.json
@@ -18,8 +18,10 @@
     "version": "pnpm version"
   },
   "dependencies": {
+    "@glimmer/syntax": "^0.84.2",
     "@rehearsal/migration-graph-shared": "workspace:*",
     "@swc/core": "^1.3.32",
+    "babel-plugin-htmlbars-inline-precompile": "^5.3.1",
     "debug": "^4.3.4",
     "enhanced-resolve": "^5.12.0",
     "fast-glob": "^3.2.12",
--- a/packages/migration-graph-ember/src/entities/ember-addon-package-graph.ts
+++ b/packages/migration-graph-ember/src/entities/ember-addon-package-graph.ts
@@ -1,7 +1,5 @@
-import fs from 'fs';
 import { resolve } from 'path';
-import { IResolveOptions } from 'dependency-cruiser';
-import { CachedInputFileSystem } from 'enhanced-resolve';
+import { type IResolveOptions } from 'dependency-cruiser';
 import debug from 'debug';
 
 import { getEmberAddonName } from '../utils/ember';
@@ -33,11 +31,9 @@ export class EmberAddonPackageGraph extends EmberAppPackageGraph {
       alias,
     });
 
-    const options = {
-      fileSystem: new CachedInputFileSystem(fs, 4000),
-      resolveDeprecations: false,
-      alias: alias,
-    };
+    const options = super.resolveOptions;
+
+    options.alias = alias;
 
     return options;
   }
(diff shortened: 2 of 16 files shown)


wondersloth:
Summary of work:

- Did not use `@glimmerx/babel-preset` or `@glimmer/babel-preset` as I want to produce some module source format that can be parsed by SWC.
- If I were to use babel to **transform** the source file using:
```
  const { code } = transformSync(output, {
    filename,
    targets: {
      node: 'current',
    },
    presets: [['@glimmerx/babel-preset', { __loadPlugins: true }]],
  });
```

It produces output that is not what we need for passing to swc to parse out the service usage. 

TL;DR I wrote the swc parser to work against es6 source, so the output from babel transform will lose that fidelity to parse the AST for service usage.

wondersloth:
Note: Prior concerns about alternative dependency graph tools, are minor because we're no longer using experimental API from dependency cruiser.

chadhietala (approved):
This looks great @wondersloth.

lynchbomb (approved):

PR #792, opened 2023-02-28
feat(cli): force migration from root dir

### LINKED ISSUE #767: force migrations from root directory only
We need a check to confirm the migration is being migrated from the project root only. A sub-package should be specified as a path flag only.

haochuan:
This fixes #767.

- Remove `--basePath` option in `rehearsal migrate`, so that `basePath` would always be where you run the CLI (`process.cwd()`). This would reduce a lot of UX confusions, and also make it easier for future dev work since the basePath or root should be super clear during the entire process. If we need to run migration against a specific package inside a workspace, we would introduce a new option `--package` to achieve this.
- Check if the current dir is the root of a project with npm/yarn/pnpm workspaces. The CLI would stop the and show warning message if it's not in the root, as well as the suggested root dir detected from the CLI.

e.g. CLI would show error when running `rehearsal migrate` in a sub package dir:

```
migrate command needs to be running at project root with workspaces.
Seems like the project root should be /var/folders/ww/workspace-test instead of current directory (/var/folders/ww/workspace-test/packages/package-a)
```
- Check if `entrypoint` is inside the current root of the project.

e.g. in dir `/code/dummy-app`, running `rehearsal migrate -e ../index.js`

```
Could not find entrypoint /code/index.js. Please make sure it is existed and inside your project(/code/dummy-app).
```


packages/cli/src/commands/migrate/index.ts
```diff
@@ -77,8 +76,8 @@ async function migrate(options: MigrateCommandOptions): Promise<void> {
   // 1. Not --dryRun
   // 2. Not --regen
   // 3. First time run (check if any report exists)
-  if (!options.dryRun && !options.regen && !reportExisted(options.basePath, options.outputPath)) {
-    const hasUncommittedFiles = await gitIsRepoDirty(options.basePath);
+  if (!options.dryRun && !options.regen && !reportExisted(basePath, options.outputPath)) {
```
egorio, line 79: Why don't keep basePath as part of *Options types with default ` = process.cwd()`?

You can use [.hideHelp()](https://www.npmjs.com/package/commander#user-content-more-configuration), or just don't expose this to CLI at all.

haochuan: I was hiding it exactly as you suggested here at first. But then thought if we just hide it and don't use it as an option, we could just remove it and take it out of `options`.

It makes more sense to me doing this way, since we agreed to make it a hard requirement to run the CLI from project root so that `basePath` won't be a variable in any case. However, if at some point we need to pass the `basePath` as option again, then for sure we need to keep it within `options`.

PR #797, opened 2023-03-02
feat: force migration from root dir

### LINKED ISSUE #767: force migrations from root directory only

haochuan:
This fixes #767.
This is a new PR to replace #792 

- Hide `basePath` option in CLI, default to `process.cwd()`, and force it to `process.cwd()` even user passes a value accidentally.

- Check if the current dir is the root of a project with npm/yarn/pnpm workspaces. The CLI would stop the and show warning message if it's not in the root, as well as the suggested root dir detected from the CLI.

e.g. CLI would show error when running `rehearsal migrate` in a sub package dir:

```
migrate command needs to be running at project root with workspaces.
Seems like the project root should be /var/folders/ww/workspace-test instead of current directory (/var/folders/ww/workspace-test/packages/package-a)
```
- Check if `entrypoint` is inside the current root of the project.

e.g. in dir `/code/dummy-app`, running `rehearsal migrate -e ../index.js`

```
Could not find entrypoint /code/index.js. Please make sure it is existed and inside your project(/code/dummy-app).
```


egorio (commented):
Thanks, @haochuan. 
I don't think we need to "ensure process.cwd() is basePath even use passes anything accidentally"... I think it's fine to allow it but don't guarantee the work since this is not a publicly-available option... but I think it's good for now to force, we can revert it later if needed. 

@lynchbomb Please review as well.

egorio (approved):

PR #805, opened 2023-03-03
feat(cli): init command and skip configs

### LINKED ISSUE #749: Ability to skip config steps
Users request to be able to skip steps when Rehearsal creates configs (for instance if they already have them).

Can be done by separating the "code fixes" step: `migrate fix`.

### LINKED ISSUE #750: Better output during bootstrap (configs, dependencies)
Users want to inspect what modifications are being made during bootstrap.
 
 CLI to output config values being added/created and the list of dependencies being installed.

### LINKED ISSUE #751: Created configs conflicting with existing project setup
This is a problem with most monorepos as they already have some level of bootstrapping.
Some monorepo apps have eslint config in root and nowhere else.

We can provide more options for creating configs command `migrate init`, like:

- [ ] `--ts=***` - configure to specific ts version to use (`stable` for `migrate`, `latestBeta` for `upgrade`)
- [ ] `--config=ts,lint,...` generate only specific config and install related deps
- [ ] `--override` - override existing configs with proposed
- [ ] `--append` (`--strictify`) - append missing options to existing configs to make them stricter 
- [ ] `--skip` - skip this step if files exist
- [ ] `--tsConfigPath` = path to ts config files (if different from --entry)
- [ ] `--eslintConfigPath` = path to eslint config files (if different from --entry)

These options can be passed through `--userConfig`

### LINKED ISSUE #779: Skip config tasks individually, not based on report existence
We currently skip config tasks if the `.rehearsal` dir exists.
This brings to error when the user is trying to `git reset` and run the command again.

https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/migrate/tasks/validate.ts#L77-L82

```
      if (reportExisted(options.basePath, options.outputPath)) {
        ctx.skipDepInstall = true;
        ctx.skipTsConfig = true;
        ctx.skipLintConfig = true;
        ctx.skipScriptConfig = true;
      }
```

Instead, we should check if each task needs to be skipped individually, for instance (conditions can be different):

- `ctx.skipDepInstall = true` if all deps we need already installed (part of package json).
- `ctx.skipTsConfig = true` if ts config exists
- ...

### Steps to reproduce:

```
mint clone ember-cli-pemberly-m3
cd ember-cli-pemberly-m3
rehearsal migrate -e addon/
git reset --hard
rehearsal migrate -e addon/
```

haochuan:
Fixes #779 
Fixes #749 
Fixes #750 
Fixes #751 

### New subcommand rehearsal migrate init

- run `depInstallTask`, `tsConfigTask`, `lintConfigTask` and `createScriptsTask`
- accept `-u --userConfig` option, which reads `migration-config.json` for custom deps, ts config and lint config

### Task updates (depInstallTask, tsConfigTask, lintConfigTask and createScriptsTask)

- Smarter `shouldRun` checks to see if it meets the prereqs and if we should run the task
- Detailed message to show what the CLI is doing (e.g. print all dependencies we are trying to install)

### rehearsal migrate updates
- Now when you run `rehearsal migrate`, it would check if the current project passes the check in `depInstallTask`, `tsConfigTask`, `lintConfigTask` and `createScriptsTask`, to decide if we need to run those tasks.
- New option `--skip-init` to force bypass `depInstallTask`, `tsConfigTask`, `lintConfigTask` and `createScriptsTask`, which is hidden to end users by default.


packages/cli/src/commands/migrate/init-command.ts
```diff
@@ -0,0 +1,49 @@
...
+  const loggerLevel = options.verbose ? 'debug' : 'info';
+  const logger = createLogger({
+    transports: [new transports.Console({ format: format.cli(), level: loggerLevel })],
+  });
+  const defaultListrOption = {
+    concurrent: false,
+    exitOnError: true,
+  };
+
+  const tasks = [
```
lynchbomb, line 35: the functionality we are looking for would make this child command atomic. we can combine the tasks in the migrate/index.ts command. for this command it shouldn't run any sibling tasks. 

`rehearsal migrate` <<< will run the tasks including init
`rehearsal migrate init` <<< will only run init

lynchbomb (changes requested):

haochuan:
Some notes and stuff to discuss:

1. What would `migrate init` do?

From a user perspective, `migrate init` would do

- depInstallTask: dependency installation
- tsConfigTask: how to handle `tsconfig.json`
- lintConfigTask: how to handle `.eslintrc`
- createScriptsTask: add additional scripts in `package.json`

Currently the existed `--init` option would run:

```
 const tasks = [
    await validateTask(options, logger),
    await initTask(options),
    await depInstallTask(options),
    await tsConfigTask(options),
    await lintConfigTask(options),
    await createScriptsTask(options),
  ];
```

2. `migrate init` vs the actual `initTask`

Current `initTask` !== what we should run in `migrate init`. `initTask` would prepare stuff for the rest tasks in the CLI, not for the end users.

3. Do we want to add option to skip one/multi tasks in `migrate init`?
4. Do we want to add option to skip `migrate init` when running `migrate`?

haochuan:
After discussion with @lynchbomb we need to:

- create new subcommand `rehearsal migrate init`
  - do dep install, ts setup, lint setup, script setup
  - each task should be smart enough to check if it should run
    - in ts setup with user config, check for strict mode
  - in dep install task, have a better UI and error handling message, to let the user know what exactly goes wrong, and what's the next action
 
- `rehearsal migarte`
   - also run `migrate init`
   - new options `--skip-init`, but hide it in help text

haochuan:
> @haochuan is this ready for review?

@lynchbomb yeah but saw a lint error, need to fix that first.

I was trying to use `typescript.parseJsonConfigFileContent` to read the extends in `tsconfig.json`, but seems import `typescript` is not allowed here, need to figure out another way

haochuan:
Removed `typescript` import and use [`get-tsconfig`](https://github.com/privatenumber/get-tsconfig) to parse extensions. All lints and tests passed in my local.

lynchbomb (approved):

lynchbomb:
Awesome work! This is a much needed workflow.

PR #808, opened 2023-03-04
chore: run smoke on branch

chadhietala:
This PR should enable us to run the smoke test against the pr branch. 

## Why
Prior to this commit we would only run the smoke test against the master branch, which would leave us in the state where master was 🔴 . This prevented us from releasing broken software but ideally we want to shift left and test our PRs against the smoke test app prior to merging. 

## What Else

- I added names to the workflow file so they look nicer
Screenshot 2023-03-05 at 7 49 57 PM
- There is no need to have the vitest config to be a `.ts` file now that we are in ESM. Master seems to be broken because this config can't be loaded.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -11,11 +11,20 @@ jobs:
     name: Runs the CLI Smoke Test
     runs-on: ubuntu-latest
     steps:
-      - uses: convictional/trigger-workflow-and-wait@v1.6.5
+      - name: Get branch name
+        uses: tj-actions/branch-names@v6
+        id: branch-name
+
+      - name: echo name
+        run: |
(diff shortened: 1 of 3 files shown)


lynchbomb:
Dude this will be slick if it works as advertised

lynchbomb:
Also we will need to update the smoke-test action https://github.com/rehearsal-js/smoke-test/blob/master/.github/workflows/run-smoke-test.yml and pass the `client_payload` into the .sh script with `branch_name`

here ---> https://github.com/rehearsal-js/smoke-test/blob/master/download-rehearsal-cli.sh#L10

we can alternatively just pass the commit sha into this script and get the zip from a different endpoint eg. https://github.com/rehearsal-js/rehearsal-js/archive/COMMIT_SHA.zip

lynchbomb (approved):

PR #809, opened 2023-03-04
refactor: remove rehearsal commits on migrate only

### LINKED ISSUE #754: Don't check for uncommitted files and create git commits by Rehearsal (by default)
Rehearsal currently creates git commits at some steps.
This is not convenient for those users who want to review/modify files before committing them.
We can change Rehearsal behavior to 
- [ ] work without committing any files and just by ignoring any uncommitted changes.
- [ ] do not create any commits during its workflow (or create an option to create / not create commits)
- [ ] create more granular command to allow users to run only steps and get a small amount if they want to

lynchbomb:
This PR fixes #754 and removes rehearsal running `git add *` for `migrate` command only


haochuan (approved):
👍

PR #814, opened 2023-03-06
test: expand migrate app fixtures

lynchbomb:
This PR expands the end-to-end cli migration to include a more robust fixture project with ESM.


packages/cli/test/commands/migrate/convert-projects.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  test('migrate from default all files .js in root', async () => {
+    const options = createMigrateOptions(basePath);
+    // Get context for convert task from previous tasks
+    const tasks = [
+      await initTask(options),
+      await depInstallTask(options),
+      await tsConfigTask(options),
+      await lintConfigTask(options),
+      await convertTask(options, logger),
```
chadhietala, line 63: These do not return promises and should not be awaited.

chadhietala (changes requested):

PR #819, opened 2023-03-07
Integrate Glint Language Service

cafreeman:
This PR integrates [glint](https://github.com/typed-ember/glint) into Rehearsal via a new `GlintService` and a set of glint-specific plugins. This enables Rehearsal to run on ember or glimmer projects and work with `hbs` and `gjs` files in addition to regular `js` files.

High-level changes:
- Expose Glint via `GlintService` that mirrors `RehearsalService` and can be used with it interchangeably
- Add `GlintCheck` and `GlintFix` plugins for applying codefixes and todos to Glint-specific file types (e.g. `hbs` and `gjs` files)
- Added a filter mechanism for plugins so that you can specify when a plugin should actually run (this allows us to make the glint plugins and rehearsal plugins mutually exclusive)
- Added logic for determining when to use `RehearsalService` vs `GlintService` based on the dependencies present in the root of the project. If we find `ember-source`, `@glimmer/component`, or `@glimmerx/component`, we use Glint.


egorio:
Looks great!

packages/migrate/test/migrate.test.ts
```diff
@@ -0,0 +1,430 @@
...
+        basePath: project.baseDir,
+        sourceFiles: inputs,
+        entrypoint: '',
+        reporter,
+        logger,
+      };
+
+      await migrate(input);
+
+      const expected = `<template>
```
chadhietala, line 111: Is there a reason why we want this here instead of using snapshots?

packages/migrate/test/migrate.test.ts
```diff
@@ -0,0 +1,430 @@
...
+/* @ts-expect-error @rehearsal TODO TS2339: Property 'name' does not exist on type 'Foo'. */
+    return this.name;
+  }
+}
+`;
+
+      expectFile(outputs[0]).toEqual(expected);
+    });
+
+    test.skip('complex', async () => {
```
chadhietala, line 352: skip?

packages/plugins/package.json
```diff
@@ -37,6 +37,7 @@
     "version": "pnpm version"
   },
   "dependencies": {
+    "@glint/core": "file:.yalc/@glint/core",
```
chadhietala, line 40: I'm assuming the dependency is just temporary? We likely want this to be a peerDep because rehearsal it self should install this into the developer's project.

packages/plugins/src/plugins/glint-fix.plugin.ts
```diff
@@ -0,0 +1,138 @@
...
+import { CodeActionKind, Diagnostic } from 'vscode-languageserver';
+
+export class GlintFixPlugin implements Plugin<PluginOptions> {
+  async run(fileName: string, context: PluginsRunnerContext): PluginResult {
+    const allFixedFiles: Set<string> = new Set();
+    const service = context.service as GlintService;
+
+    const diagnostics = this.getDiagnostics(fileName, service);
+
+    for (const diagnostic of diagnostics) {
```
chadhietala, line 19: I'm kind of surprised that you don't run into bugs here with looping over the diagnostics and applying changes in the loop. Since we use `diagnostic.range` to compute where we need to splice in the codefix the `range` immediately becomes stale on the next turn of the loop. I think you need to recall `this.getDiagnostics` and re-assign `diagnostics` at the end to ensure you're always getting the new text ranges.

packages/plugins/src/plugins/glint-fix.plugin.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    return service.transformCodeActionToCodeFixAction(actions);
+  }
+
+  private makeCodeFixStrict(fix: CodeFixAction): CodeFixAction | undefined {
+    // Filtering out all text changes contain `any`
+    const safeChanges: FileTextChanges[] = [];
+    for (const changes of fix.changes) {
+      const safeTextChanges: TextChange[] = [];
+      for (const textChanges of changes.textChanges) {
```
chadhietala, line 100: We should try to consolidate this logic somehow so that we don't have two copies of it in the codebase.

packages/migrate/test/migrate.test.ts
```diff
@@ -0,0 +1,430 @@
...
+    const ext = extLookup[fileExt];
+
+    return changeExtension(file, ext);
+  });
+
+  return [inputs, outputs];
+}
+
+describe('migrate', () => {
+  describe('.gjs', () => {
```
chadhietala, line 67: Can we add some cases where we have a `class` based implementation?

chadhietala (changes requested):
Overall looks great. Most of my feedback are clarifying questions / observations but I do think there is a bug in the glint.fix.plugin. The best way to repro it is to have 2 errors in the file and where after applying the fix from first it will be in the text range of the other.

packages/migrate/test/migrate.test.ts
lynchbomb: I looked into this on Saturday. This was the test skipped in the ESM migration. The issue is the way this test is architected. UUID and importer types are being used elsewhere in the monorepo and the types are already available. As such the updated package.json devDep will never be included (since it already is). 

Additionally in debugging (can’t 100% validate but is replicated) if we specify a different module which doesn’t bundle types and we don’t already consume it. TypeScript 4.9 is auto pulling in the correlated @types in the typescript global cache.

packages/plugins/package.json
cafreeman: yea I can move this to peer/devDeps before we merge

packages/migrate/test/migrate.test.ts
cafreeman: There's a `with-class` case in this section that does have a class-based component with an inlined `<template>`. Is that what you mean?

packages/migrate/test/migrate.test.ts
lynchbomb: There are a lot of these. @cafreeman if snapshots aren't ideal; what if we just move these to static .output test files?

lynchbomb:
Generally this is awesome. Excellent work! Beyond what is flagged by @chadhietala, merge conflicts and nit around inline expected test output this is good to go (once above is fixed)

packages/plugins/src/plugins/glint-fix.plugin.ts
cafreeman: Updated this to behave like the `diagnostic-fix` plugin (which does what you suggested)

packages/migrate/test/migrate.test.ts
cafreeman: @lynchbomb I'm not opposed to moving to snapshot tests if we really want to do that, but I honestly think having the inlined expected outputs is nice since it makes it *very* easy to see what `migrate` is actually doing.

packages/plugins/package.json
cafreeman: moved

packages/plugins/src/plugins/glint-fix.plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+  getDiagnostics(service: GlintService, fileName: string): Diagnostic[] {
+    return service.getDiagnostics(fileName).map((d) => service.convertTsDiagnosticToLSP(d));
+  }
+
+  getCodeFix(
+    fileName: string,
+    diagnostic: Diagnostic,
+    service: GlintService
+  ): CodeFixAction | undefined {
+    const glintService = service.getGlintService();
```
egorio, line 71: Your work looks awesome! 

But I'm curious why does diagnostic plugin was not used directly?
The logic of converting Glint codefixes to TS codefixes cab be hidden inside service.getDiagnostics + logic to applying them can be moved to both services as well, so:

- rehearsal.getCodeFix and gling.getCodeFixes (includes positionToOffset) return the same format
- diagnostic-fix just called geDiagnostic -> getCodeFixes -> applies the same way for both services

egorio (approved):

packages/plugins/src/plugins/glint-fix.plugin.ts
cafreeman: Good question. Given the subtle differences between applying codefixes from Glint vs from Rehearsal it felt more straightforward to just separate them and have a separate plugin for gjs/hbs files rather than trying to conditionally handle everything in one plugin.

chadhietala (approved):

PR #829, opened 2023-03-08
feat(cli): migrate default interactive - optional ci

lynchbomb:
This PR makes `--interactive` the default and removes the need for the flag. Additionally it adds a new flag `--ci` which is the inverse of interactive.

@chadhietala `--ci` will need to be updated for all automated scripts / actions

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,11 +73,11 @@ Rehearsal Migrate has performed the following tasks:
 
 ## Interactive Mode - Migrate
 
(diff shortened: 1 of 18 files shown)


egorio:
👍 but I think we need to remove `--ci` as well in the future and just pass all required answers to any interactive questions as CLI options.

lynchbomb:
@egorio yes and/or from the config

egorio (approved):

haochuan (approved):
Looks great!

PR #837, opened 2023-03-10
wip: diagnostic and codefix reproducability

chadhietala:
This PR is going to help bring reproducibility of the codefix portion of the `migrate`. In addition it will:

1. Generate Diagnostic Message objects directly from the TypeScript codebase
2. Generate CodeFix Message objects which represent all the cases we `migrate` supports
3. Enumerate `passing` and `failing` examples that the codefixes actually fix and use it to generate tests
4. By having a place to enumerate the examples we have an a singular place to reproduce issues
5. Limited the codefixes to only the diagnostics associated with those codefixes
6. Moved to async generators all the way through into the internal loops of the plugins to allow us to pass in state in the future

This is still a WIP but I'm getting it off my machine


.github/workflows/download-typescript.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  push:
+    branches: [master]
+
+jobs:
+  get_diagnostics:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download Diagnostics
+        run: curl https://raw.githubusercontent.com/microsoft/TypeScript/main/src/compiler/diagnosticMessages.json -o packages/codefixes/src/diagnosticMessages.json
```
egorio, line 12: I don't think it makes sense to copy this file. I think it's better to download it from TS repo because it's updated frequently.

packages/codefixes/src/codefixInformationMap.generated.ts
```diff
@@ -0,0 +1,461 @@
...
+  failing: string[]
+): CodefixMessage {
+  return { name, title, description, passing, failing, diagnostics };
+}
+
+/** @internal */
+export const CodeFixes = {
+  addMissingAsync: fix(
+    'addMissingAsync',
+    'Add Missing `async` Keyword',
```
egorio, line 30: Why do we need these messages to "copy" to our repo if we always have access to then when they are available to apply?

packages/codefixes/src/supported-codefixes.ts
```diff
@@ -0,0 +1,59 @@
+import { DiagnosticWithLocation } from 'typescript';
+import { CodeFixes } from './codefixInformationMap.generated.js';
+import { Diagnostics } from './diagnosticInformationMap.generated.js';
+
+const PRIORITY_DIAGNOSTICS = [
+  Diagnostics.This_constructor_function_may_be_converted_to_a_class_declaration.code,
```
egorio, line 6: I found `Diagnostics.***.code` way harder to work with than working with codes directly. 

As an engineer, you always know the error code, but reading and debugging those constants is annoying. 
Ts uses it because they have translations and those values are used at keys, we don't need it.

image


image

packages/codefixes/src/codefixesMessages.json
```diff
@@ -0,0 +1,271 @@
+{
+  "addMissingAsync": {
+    "title": "Add Missing `async` Keyword",
```
egorio, line 3: I would keep this array without converting it to ".generated" with `Diagnostic.***.code`, we don't need them in Rehearsal.

.github/workflows/download-typescript.yml
egorio: But I think we don't need this step at all :)

packages/codefixes/src/supported-codefixes.ts
egorio: This list is simpler to consume like this (because you can match the error code with exact error you see code in rehearsal comment, IDE or build output) without "converting" it to some constant and do the back when debugging:

```
prioritizedCodes = [
    80002, 
    80005, 
    ...
    7005, 
    ...
]
```

You can refer to `[codefixesMessages.json](https://github.com/rehearsal-js/rehearsal-js/pull/837/files#diff-dc581959aff8155e493b0c6c12c7bb3837c640a10b0c81d8ee662d6b630d5590)` if need info about error.

I had a hard time debugging TS because of those `Diagnostic.*` constants, would be happy to avoid this in Rehearsal.

egorio:
@chadhietala don't think we need 1 and 2 in Rehearsal. I have a hard time working with those in TS.

Kaceykaso:
> @chadhietala don't think we need 1 and 2 in Rehearsal. I have a hard time working with those in TS.

It looks like this change, while still largely a WIP, is about making Rehearsal easier to debug, and easier for the consumer to reason about the error. Converting error codes to more user friendly error messages is something ALL of our tools should do. Also it seems this PR is also taking a stab at some e2e tests, which Rehearsal needs.  Unless you have some solid data points from early adopter feedback, but I would say our feedback is still small and simple.

egorio:
@Kaceykaso

>  Converting error codes to more user friendly error messages is something ALL of our tools should do

We already cover codes to messages for end users, this PR converts "variable names/values" to "error messages"  :(
Which is very painful to work with. I spent days to debug TS code with such variables because it's very hard to handle by humans. 

In the case of rehearsal engineers work directly around error codes: error codes are part of `@rehearsal` messages and `-build` output. If you need to debug, add or find a fix, etc. you always know the error code, but not its generated constant name like `Parameter_0_implicitly_has_an_1_type_but_a_better_type_may_be_inferred_from_usage` where 0 and 1 represent some real names. 

Another reason, it's very hard to read through the code like this, is because texts are very similar:

```
Diagnostics.Parameter_0_implicitly_has_an_1_type_but_a_better_type_may_be_inferred_from_usage
    .code,
 Diagnostics.Member_0_implicitly_has_an_1_type_but_a_better_type_may_be_inferred_from_usage.code,
 Diagnostics
    .Variable_0_implicitly_has_type_1_in_some_locations_but_a_better_type_may_be_inferred_from_usage
    .code,
 Diagnostics._0_implicitly_has_an_1_return_type_but_a_better_type_may_be_inferred_from_usage.code,
  Diagnostics
```
instead of this:
```
7044
7045
7046
7050
```

> Unless you have some solid data points from early adopter feedback

There is nothing changed for early adopters by this PR.

chadhietala:
Superceded by #844

PR #844, opened 2023-03-13
Operationalize TypeScript Codefixes

chadhietala:
This PR operationalizes how we add new codefixes coming from the TypeScript Language Service. It does so by creating a manifest of all the supported codefixes and their associated diagnostic codes to do 3 things:

1. Creates a workflow for onboarding new codefixes
2. Creates minimal testing of the codefixes to ensure they can be targeted correctly
3. [Generates documentation](https://github.com/rehearsal-js/rehearsal-js/blob/a037c9ac1583768785e7717b483ef8d4190163b3/Supported-Fixes.md) from the tests so consumers know what they can expect
4. Constrains transformation of code to only the things in the manifest to create better reproducibility and debugging.


Supported-Fixes.md
```diff
@@ -0,0 +1,812 @@
+# Supported CodeFixes
```
cafreeman, line 1: 👏 👏 👏 👏 👏 👏

cafreeman (approved):

lynchbomb (approved):
Excellent

Supported-Fixes.md
lynchbomb: This is fantastic and we can leverage this within PR’s for upgrade

packages/codefixes/src/diagnosticMessages.json
```diff
@@ -0,0 +1,7744 @@
+{
```
egorio, line 1: .gitignore this file?

packages/codefixes/src/codefixes.ts
```diff
@@ -1,26 +1,54 @@
...
-    2790: new MakeMemberOptionalCodeFix(),
-    4082: new AddMissingExportCodeFix(),
-    7050: new AddMissingTypesBasedOnInlayHintsCodeFix(),
+    [Diagnostics._0_is_of_type_unknown.code]: new AddErrorTypeGuardCodeFix(),
+    [Diagnostics.Object_is_of_type_unknown.code]: new AddErrorTypeGuardCodeFix(),
+    [Diagnostics.The_operand_of_a_delete_operator_must_be_optional.code]:
+      new MakeMemberOptionalCodeFix(),
+    [Diagnostics.Default_export_of_the_module_has_or_is_using_private_name_0.code]:
+      new AddMissingExportCodeFix(),
+    [Diagnostics._0_implicitly_has_an_1_return_type_but_a_better_type_may_be_inferred_from_usage
```
egorio, line 26: I think `[7050]: [new AddMissingTypesBasedOnInlayHintsCodeFix()]` is easy to work with. 

You can easily overview which error code is associated with a particular fix, you don't need to read through messages or do additional "decoding" steps to recognize when an error code is associated.

packages/codefixes/src/safe-codefixes.ts
```diff
@@ -0,0 +1,9 @@
+import { CodeFixes } from './codefixInformationMap.generated.js';
+
+export const isCodeFixSafe = (fixId: string): boolean => {
```
egorio, line 3: Seems can be renamed to `isCodeFixSupported`

packages/codefixes/test/codefixes.test.ts
```diff
@@ -0,0 +1,229 @@
...
+            service.setFileText(filename, text);
+          },
+        });
+      }
+    }
+
+    expect(service.getFileText(input)).toEqual(await fs.readFile(expectation, 'utf-8'));
+  }
+
+  test('addMissingAync', async () => {
```
egorio, line 85: You can use `test.each` or even `describe.each` instead of create test for every codefix
```
const test = {
  'convertToTypeOnlyImport': [
    'convertToTypeOnlyImport/failing/fail-1.ts',
    'convertToTypeOnlyImport/passing/pass-1.ts',
     ['convertToTypeOnlyImport/sample-1.ts']
  ],,

test.each(tests)(...
```

egorio: You can even generate `const codefixesTests =` from files automatically... through it looks too magic :)

egorio:
👍

PR #845, opened 2023-03-14
chore: cleanup async issues in cli

chadhietala:
This is start of the cleanup to fix all the async issues and places where `any` types have leaked into the codebase.

In future PRs we need to expand the packages and files that we use "[recommended-using-type-checking](https://typescript-eslint.io/linting/configs#recommended-requiring-type-checking)"

I also removed the eslint-recommended because the recommended config inherits from it.

https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.ts

## What's Going On Here

We were marking things like tasks as `async` functions but they are actually sync functions that return a task object that has an async task method. With the introduction to this lint config it will point to all of these issues.

Screenshot 2023-03-14 at 1 03 57 PM

Screenshot 2023-03-14 at 1 04 51 PM

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -19,7 +19,7 @@ module.exports = {
   parserOptions: {
     sourceType: 'module',
     ecmaVersion: 2022,
-    project: ['./packages/*/tsconfig.json', './packages/*/test/tsconfig.json'],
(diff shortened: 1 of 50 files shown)


egorio:
Why are we removing all async functions?

chadhietala:
@egorio Does my description answer your question.

brendaHuang (approved):

PR #846, opened 2023-03-14
chore: add fixtures to package

brendaHuang:
Package fixtures for @rehearsal/test-support

### DIFF
--- a/packages/test-support/package.json
+++ b/packages/test-support/package.json
@@ -14,7 +14,7 @@
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
   "files": [
-    "dist"
+    "dist", "fixtures", "scripts"
   ],
   "scripts": {
     "build": "tsc -b",


wondersloth (approved):

egorio:
This won't work. The release process will just be stuck trying to pack all fixtures and their node_modules.

egorio (changes requested):
We removed fixtures at some point because packages are not able to be released with them because of the size or node_modules https://github.com/rehearsal-js/rehearsal-js/pull/723/files.

If you want to return fixtures, you need to add a step to release.js to remove them. Here is a solution: https://github.com/rehearsal-js/rehearsal-js/commit/85ecd85e1342b59506d3632a8b4ac8d72bbed448

egorio:
If you want to return fixtures, you need to add a step to release.js to remove:

```
packages/test-support/fixtures/ember/addon-template/node_modules
packages/test-support/fixtures/ember/app-template/node_modules
```

wondersloth:
We possibly can add a prepack hook to the package so that it removes node_modules as well. That may help.

egorio:
@wondersloth we can just clean those folders during release: https://github.com/rehearsal-js/rehearsal-js/commit/85ecd85e1342b59506d3632a8b4ac8d72bbed448, but the question is do we need to install those deps at all?

chadhietala:
@brendaHuang what issue is this associated with?

brendaHuang:
@chadhietala I am trying to import these fixtures to be used for smoke-test.

chadhietala:
based on our sync I'm going to close this PR

PR #850, opened 2023-03-16
Strict Linting

chadhietala:
This PR fixes all\* the linting issues when we turn on the[recommended-requiring-type-checking](https://typescript-eslint.io/linting/configs#recommended-requiring-type-checking).

## Why

These lint rules add even more strictness to how to write TypeScript code by making sure that all async code is properly handled and that no `any`s enter the codebase. This will insure that refactors go much smoother and we can be confident in those changes.

## What Else

There has been some more simplification done in the migration graph area that needed to be dealt with because the explicit casting. In addition the synthetic node has been replaced with simply marking the node as `synthetic=true` and not supplying a `pkg`. This allows for us to remove a lot of the getters and setters in the entities and simply treat them more like a struct.


lynchbomb (approved):
Generally this is great. I think we can move away from type casting packageJson and Report. But not a big deal.

PR #854, opened 2023-03-16
chore: test nightly

chadhietala:
This adds a github action job to run the entire test suite against nightly builds of TypeScript. Ideally in the future we should use rehearsal it self to do this but this will have to work until then.

This will allows us to get ahead of any breakage since we generously use the compiler api.

### DIFF
--- a/.github/workflows/test-and-lint.yml
+++ b/.github/workflows/test-and-lint.yml
@@ -11,14 +11,55 @@ jobs:
     name: Runs Tests with Linting
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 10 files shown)


lynchbomb (approved):
Calling `pnpm test` twice. Otherwise LGTM; and yes we should replace this with `rehearsal upgrade -b next --dry-run`

lynchbomb:
Looks like that failing test is doing a deepEqual comparison and there's whitespace delta

lynchbomb:
This PR can be simplified from fixes which landed after it. Closing in lieu of #888

PR #857, opened 2023-03-17
chore: enable pnpm caching in github action

brendaHuang:
To speed up scenario testing. Followed the direction of [this page](https://github.com/pnpm/action-setup#use-cache-to-reduce-installation-time). Changed the output of the store_output variable from what the instruction says, to avoid a pnpm bug: pnpm store will return a multiline string for path.

### DIFF
--- a/.github/workflows/test-and-lint.yml
+++ b/.github/workflows/test-and-lint.yml
@@ -28,8 +28,10 @@ jobs:
         id: pnpm-cache
         shell: bash
         run: |
-          PNPM_STORE=$(pnpm store path)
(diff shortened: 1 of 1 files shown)


egorio (approved):

wondersloth:
I'm not sure if this will work for the scenario test explicitly because they don't use pnpm they use yarn. All the scenarios are out of the box ember apps and adding that use yarn.

wondersloth:
Talked about this briefly today at our no-sync meeting at 930AM PST.

Adding the PNPM cache is good, as it will help with CI runs and updates to PRs.
Adding the NPM cache is good, as it will help with all scenario-tests that invoke npm install.

@brendaHuang all my concerns have been addressed.

PR #858, opened 2023-03-20
feat: filtering out unavailable types coming from JSDocs

egorio:
In progress... thinking if this can be done more optimized way :(


packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -127,6 +191,22 @@ export class TypescriptCodeFixCollection implements CodeFixCollection {
...
+    let previousNode: ts.Node = sourceFile;
+
+    const visitor = (node: Node): Node | undefined => {
+      if (node.getStart() >= pos) {
+        return previousNode;
+      }
+
+      previousNode = node;
+
+      return ts.forEachChild(node, visitor);
```
egorio, line 204: Don't need to visit child node if the current node ends before pos

PR #859, opened 2023-03-20
fix: use PAS token

chadhietala:
I believe we need to pass in this PAS token so we don't get [protect brach errors](https://github.com/rehearsal-js/rehearsal-js/actions/runs/4462920273/jobs/7837784315). @lynchbomb is there anything special with permissions to use this token?


lynchbomb:
This should work fine. That token has access to read/write within smoke-test and this repo

lynchbomb:
IIRC I did give it access to branches outside master. If not we can create another which has it.

PR #862, opened 2023-03-20
chore: bumps typescript@latest

lynchbomb:
This PR bumps to TypeScript@5 (latest)


Supported-Fixes.md
```diff
@@ -370,7 +370,7 @@ console.log(b, c);
 **Output:**
 
 ```ts
-import type { B, C } from '../sample-1.js';
+  import type { B, C } from '../sample-1.js';
```
lynchbomb, line 373: Looks like the typescript codefix `convertToTypeOnlyImport` is adding this tab whitespace. since we dont run prettier for this and its the only codefix doing this, im not sure the best approach to resolve. for now i've included the tab whitespace. 

we could possibly trim whitespace from only the first line of the `fail.ts` files? @chadhietala

PR #865, opened 2023-03-20
feat: handle fully qualified service injections

cafreeman:
Implements https://github.com/rehearsal-js/rehearsal-js/issues/390


packages/migrate/src/migrate.ts
```diff
@@ -101,13 +86,13 @@ export async function* migrate(input: MigrateInput): AsyncGenerator<string> {
...
   );
 
-  const { options, fileNames: someFiles } = parseJsonConfigFileContent(
-    config,
-    ts.sys,
-    dirname(configFile),
-    {},
-    configFile
-  );
+  const parsed = parseJsonConfigFileContent(config, ts.sys, dirname(configFile), {}, configFile);
```
egorio, line 89: @cafreeman  can we move this outside of migrating the package to keep it independent from any framework-specific details?

packages/plugins/src/plugins/index.ts
```diff
@@ -5,3 +5,4 @@ export { PrettierPlugin, PrettierPluginOptions } from './prettier.plugin.js';
 export { ReRehearsePlugin, ReRehearsePluginOptions } from './re-rehearse.plugin.js';
 export { GlintCheckPlugin } from './glint-check.plugin.js';
 export { GlintFixPlugin } from './glint-fix.plugin.js';
+export { TransformPlugin } from './transform.plugin.js';
```
egorio, line 8: I think something like `ServiceInjectionsTransformPlugin` would be better :)

cafreeman: totally agree

packages/migrate/src/migrate.ts
cafreeman: This actually existed prior to the glint stuff, but I did move all the glint-specific functions to a `glint-utils` file like we discussed

packages/migrate/test/migrate.test.ts
```diff
@@ -166,6 +166,39 @@ export default class Hello extends Component {
       expectFile(outputs[0]).toEqual(expected);
     });
 
+    test('with service', async () => {
```
chadhietala, line 169: @cafreeman I know the task is to deal with namespaced services but can you add a test case for non namespaced services to make sure what we get as output is sane?

lynchbomb (approved):

PR #870, opened 2023-03-21
test: add ember tests for config-ts

brendaHuang-2023:
add ember tests for config-ts and add clean up

### DIFF
--- a/packages/cli/test/commands/migrate/__snapshots__/config-lint.ember-app.test.ts.snap
+++ b/packages/cli/test/commands/migrate/__snapshots__/config-lint.ember-app.test.ts.snap
@@ -216,3 +216,127 @@ exports[`Task: config-lint -- ember-app-matrix > ember3-appWithInRepoEngine > ru
 [SUCCESS] Create eslint config
 "
 `;
+
+exports[`Task: config-lint for ember app > ember4App > create .eslintrc.js if not existed 1`] = `
+"[STARTED] Install dependencies
+[DATA] Installing devDependencies: @types/node, @typescript-eslint/eslint-plugin, @typescript-eslint/parser, eslint, eslint-config-prettier, eslint-plugin-prettier, prettier, typescript
+[SUCCESS] Install dependencies
+[STARTED] Create eslint config
+[DATA] Create .eslintrc.js, extending Rehearsal default eslint-related config
+[SUCCESS] Create eslint config
+"
+`;
+
+exports[`Task: config-lint for ember app > ember4App > postLintSetup hook from user config 1`] = `
+"[STARTED] Install dependencies
+[DATA] Installing devDependencies: @types/node, @typescript-eslint/eslint-plugin, @typescript-eslint/parser, eslint, eslint-config-prettier, eslint-plugin-prettier, prettier, typescript
+[SUCCESS] Install dependencies
+[STARTED] Create eslint config
+[DATA] Create .eslintrc.js from config
+[DATA] Run postLintSetup from config
+[SUCCESS] Create eslint config
+"
+`;
+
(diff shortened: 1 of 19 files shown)


packages/cli/test/commands/migrate/config-ts.ember-addon.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+describe('Task: config-ts ember app', () => {
+  addonScenarios.forEachScenario((scenario: Scenario) => {
+    describe(scenario.name, () => {
+      let app: PreparedApp;
+      let output = '';
+
+      beforeEach(async () => {
+        app = await scenario.prepare();
+        clean(app.dir);
```
lynchbomb, line 24: this is cleaning beforeEach, however we need to clean afterEach instead (since the last test execution wont be cleaned with this implementation)

lynchbomb (changes requested):
left some comments

packages/cli/test/commands/migrate/config-ts.ember-addon.test.ts
brendaHuang-2023: switched to project

packages/cli/test/commands/migrate/validate.test.ts
```diff
@@ -13,6 +14,8 @@ import {
 } from '../../test-helpers/index.js';
 import type { Project } from 'fixturify-project';
 
+setGracefulCleanup();
```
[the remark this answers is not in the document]
brendaHuang-2023: removed

packages/cli/test/commands/migrate/__snapshots__/config-lint.ember-app.test.ts.snap
```diff
@@ -216,3 +216,127 @@ exports[`Task: config-lint -- ember-app-matrix > ember3-appWithInRepoEngine > ru
...
+[DATA] Installing devDependencies: @types/node, @typescript-eslint/eslint-plugin, @typescript-eslint/parser, eslint, eslint-config-prettier, eslint-plugin-prettier, prettier, typescript
+[SUCCESS] Install dependencies
+[STARTED] Create eslint config
+[DATA] Create .eslintrc.js from config
+[DATA] Run postLintSetup from config
+[SUCCESS] Create eslint config
+"
+`;
+
+exports[`Task: config-lint for ember app > ember4App > run custom config command with user config provided 1`] = `
```
lynchbomb, line 241: this snapshot seems incorrect for this test. also it seems we can do all of these snapshot checks with 1 snapshot. the delta between all of these is minimal. lets run the entire process and assert against the final snapshot instead of doing it in tiny iterations.

lynchbomb (changes requested):
At the very least for both https://github.com/rehearsal-js/rehearsal-js/pull/870/files#diff-ffd33e9cfc29f19723e0a48bdb2b82ca948ce0d7831802b14b42304384c20d29 and https://github.com/rehearsal-js/rehearsal-js/pull/870/files#diff-b0f7fe77ec2c18338e57c875bbaac7eb0e5b7631fb74df4f2c589fdb6123e4ad its confusing and adds unnecessary complexity to snapshot each tiny variation of the `stdout` instead just run the execution to completion and assert against the snapshot in its entirety. 

only when you must have another snapshot because of disparity should you generate one.

lynchbomb:
Also I dont understand why this PR doesn't have GH Action checks? https://github.com/rehearsal-js/rehearsal-js/pull/870/checks. This isn't anything you've done, just wacky.

brendaHuang-2023:
There are too many changes in too many files in this PR. I am going to separate them into smaller PRs.

PR #872, opened 2023-03-22
test: enable ember4 for ember app

brendaHuang-2023:
Enable ember 4 for ember app testing.

### DIFF
--- a/packages/cli/test/commands/migrate/init-command.ember-app.test.ts
+++ b/packages/cli/test/commands/migrate/init-command.ember-app.test.ts
@@ -1,10 +1,14 @@
 import { resolve } from 'node:path';
 import { readdirSync } from 'node:fs';
-import { beforeEach, afterEach, describe, test, expect } from 'vitest';
-import { Scenario, PreparedApp } from 'scenario-tester';
+import { beforeEach, afterEach, afterAll, describe, test, expect } from 'vitest';
 import { writeJSONSync } from 'fs-extra';
-import { appScenarios, clean, getEmber4AppProject } from '@rehearsal/test-support';
 import { Project } from 'fixturify-project';
+import {
+  getEmberAppProject,
+  getEmberAppWithInRepoAddonProject,
+  getEmberAppWithInRepoEngineProject,
+  getEmber4AppProject,
+} from '@rehearsal/test-support';
 import { REQUIRED_DEPENDENCIES } from '../../../src/commands/migrate/tasks/dependency-install.js';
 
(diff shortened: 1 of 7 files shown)


packages/test-support/package.json
```diff
@@ -41,6 +41,9 @@
     "tmp": "^0.2.1"
   },
   "devDependencies": {
+    "ember-cli-4.4": "npm:ember-cli@~4.4.0",
```
lynchbomb, line 44: why do we need to specify npm here?
eg. 
```
"ember-cli": "^4.4.0",
"ember-data": "^4.4.0",
"ember-source": "^4.4.0"
```

packages/test-support/src/project.ts
```diff
@@ -174,6 +174,31 @@ export function emberAppTemplate(): Project {
...
+  const importRoot = dirname(fileURLToPath(import.meta.url));
+  const result = importRoot.match(/.+test-support/);
+  if (!result || !Array.isArray(result)) {
+    return project;
+  }
+  const testSupport = result[0];
+
+  project.linkDevDependency('ember-source', {
+    baseDir: testSupport,
+    resolveName: 'ember-source-4.4',
```
lynchbomb, line 189: if you change above these will have a different resolve name to match

packages/cli/test/commands/migrate/init-command.ember-app.test.ts
```diff
@@ -80,3 +81,60 @@ describe('migrate init for ember app variant', () => {
...
+    await project.write();
+    basePath = project.baseDir;
+  });
+
+  afterEach(() => {
+    project.dispose();
+    basePath = '';
+  });
+
+  test(`default run -- ember 4 app`, async () => {
```
lynchbomb, line 100: these tests also need to inspect every new .ts file and confirm the type inference is valid. this can be done with snapshots.

lynchbomb (changes requested):
added some comments

packages/test-support/package.json
brendaHuang-2023: Changed

packages/test-support/src/project.ts
brendaHuang-2023: Changed

packages/cli/test/commands/migrate/init-command.ember-app.test.ts
brendaHuang-2023: This are the tests for migrate init, not migrate. There's no type inference yet at this stage.

lynchbomb (approved):

PR #874, opened 2023-03-23
fix: use service to read file text for diagnostics

cafreeman:
Should fix the issue where `migrate` was stalling out when converting TS/HBS file pairs.

When running the `diagnostic-check` plugin, we'd get the file's contents by calling `diagnostic.file.getFullText()`. In regular JS files, this is fine, but when you have a `gjs` file or `ts`/`hbs` pair, `getFullText` returns the Glint-transformed version of the file (which is basically unusable to the rest of rehearsal).

To handle this, we can instead use `service.getFileText` and pass in the filename from `diagnostic.file`.

### DIFF
--- a/packages/migrate/test/fixtures/project/src/salutation.hbs
+++ b/packages/migrate/test/fixtures/project/src/salutation.hbs
@@ -1,0 +1,1 @@
+<span>Hello {{this.name}}</span>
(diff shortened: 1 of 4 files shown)


brendaHuang-2023 (approved):
Test passed with this commit.

lynchbomb (approved):

PR #878, opened 2023-03-24
chore: fix master

chadhietala:
Fixes master branch. #856 was merged with existing failures. This fixes the failing tests by regenerating the snapshots and adding the correct config in the tsconfig.json.

PR #880, opened 2023-03-24
fix: adjust test results

brendaHuang-2023:
This test seems to fail on local and in pipeline.


brendaHuang-2023:
Close this PR since there's one already: https://github.com/rehearsal-js/rehearsal-js/pull/878

PR #881, opened 2023-03-24
test: validate ember tests

brendaHuang-2023:
ember tests for validate


packages/cli/test/commands/migrate/validate.test.ts
```diff
@@ -19,6 +17,7 @@ const logger = createLogger({
 
 describe('Task: validate', () => {
   let project: Project;
+  let basePath = '';
```
lynchbomb, line 20: you’ll need to include an afterEach hook and cleanup the project

packages/cli/test/commands/migrate/validate.test.ts
```diff
@@ -53,51 +53,35 @@ describe('Task: validate', () => {
...
   });
 
   test('error if no package.json', async () => {
-    const options = createMigrateOptions(project.baseDir);
-    const tasks = [validateTask(options, logger)];
-
-    rmSync(resolve(project.baseDir, 'package.json'));
-
-    await expect(() => listrTaskRunner(tasks)).rejects.toThrowError(`package.json does not exists`);
+    rmSync(resolve(basePath, 'package.json'));
```
lynchbomb, line 61: fixurify has an API for tmp file creation you should leverage

packages/cli/test/commands/migrate/validate.test.ts
```diff
@@ -53,51 +53,35 @@ describe('Task: validate', () => {
...
-    await listrTaskRunner(tasks);
+    await runValidate(basePath, logger, { regen: true });
     expect(cleanOutput(output, project.baseDir)).toMatchSnapshot();
   });
 
   test('pass with all config files in --regen', async () => {
-    const options = createMigrateOptions(project.baseDir, { regen: true });
-    const tasks = [validateTask(options, logger)];
-
     createFileSync(resolve(project.baseDir, '.eslintrc.js'));
```
lynchbomb, line 81: same here use fixturfy api

lynchbomb (changes requested):

packages/cli/test/commands/migrate/validate.test.ts
```diff
@@ -53,51 +53,35 @@ describe('Task: validate', () => {
   });
```
brendaHuang-2023, line 53: project.dispose() is called here. Somehow the afterEach block is collapsed in git diff.

packages/cli/test/commands/migrate/validate.test.ts
lynchbomb: delete project.files['package.json']

lynchbomb:
left a few comments, once those are resolved this should be good to merge

lynchbomb:
depending on how long `validate.ember-addon.test.ts` and `validate.ember-app.test.ts` take to run we should move these into `test:slow` and out of `test` by adding them to this exclude: https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/vitest.config.ts#L9-L16

packages/cli/test/commands/migrate/validate.test.ts
brendaHuang-2023: Tested on my local. delete project.files['package.json'] won't work, since `package.json` is automatically generated by project.write(). `package.json` is not listed in the project.files array.

packages/cli/test/commands/migrate/validate.test.ts
brendaHuang-2023: changed

brendaHuang-2023:
@lynchbomb Slightly more than 1 second to run `validate.ember-addon.test.ts`, and 1 minute to run `validate.ember-app.test.ts`. Maybe we can keep them in the original regular flow, `test`.

packages/cli/test/commands/migrate/validate.test.ts
lynchbomb: can you just delete AFTER project.write()

lynchbomb (approved):

PR #882, opened 2023-03-24
fix(migration-graph): look for service impl be either .js or .ts

### LINKED ISSUE #822: bug: when resolving a service usage to implementation, we should check to see if a ts file exists.
Internal issue, with voyager:

```
Error: Unexpected error when parsing /some-project/packages/some-web/app/adapters/-ember-m3.js. Attempting to resolve service "global-http-headers" from module/package name "global-services" in package: /some-project/packages/addons/global-services.
```
Reason it failed was that the package `global-services` impl for `global-http-headers` was migrated to `.ts` and failed the lookup on `/some-project/packages/addons/global-services/services/global-http-headers.js`

wondersloth:
This fixes #822 
- There are two scenarios, we are looking within an app, or within an addon.
- For either service implementation we have to see if there is a file with a `.ts` or a `.js` extension on disk.

### DIFF
--- a/packages/migration-graph-ember/src/entities/ember-app-package-graph.ts
+++ b/packages/migration-graph-ember/src/entities/ember-app-package-graph.ts
@@ -93,23 +93,6 @@ export class EmberAppPackageGraph extends PackageGraph {
     this.debug('>>> SERVICES DISCOVERD', services);
 
     services.forEach((s) => {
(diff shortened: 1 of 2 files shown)


wondersloth (commented):
Ready for review.

lynchbomb (approved):

Issue #876, opened 2023-03-24
[Internal] rehearsal migrate, when ts-config.json has a comment fails parsing


wondersloth:
When a `ts-config.json` exists in a project and has a comment, it breaks a json parser in rehearsal. We need this to support comments for internal products.

```
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "extends": "@linkedin/ts-bases/node16/tsconfig.json",
  "compilerOptions": {
    "allowJs": true
  }
  /**

   You will need to add relative paths in the "include" options that
   point to the root of your source and tests.

  "include": [
    //
    // For example:
    //
    // "src",
    // "test"
  ]

  */
}
```
This happening after:

```
  rehearsal:migration-graph-sahred:PackageGraph       couldNotResolve: false,
  rehearsal:migration-graph-sahred:PackageGraph       dependencyTypes: [Array],
  rehearsal:migration-graph-sahred:PackageGraph       matchesDoNotFollow: false,
  rehearsal:migration-graph-sahred:PackageGraph       circular: false,
  rehearsal:migration-graph-sahred:PackageGraph       valid: true
  rehearsal:migration-graph-sahred:PackageGraph     }
  rehearsal:migration-graph-sahred:PackageGraph   ],
  rehearsal:migration-graph-sahred:PackageGraph   dependents: [],
  rehearsal:migration-graph-sahred:PackageGraph   orphan: false,
  rehearsal:migration-graph-sahred:PackageGraph   valid: true
  rehearsal:migration-graph-sahred:PackageGraph } +1ms
  rehearsal:migrate:analyze migrationOrder [
✔ Validate project
✔ Initialize
↓ Install dependencies [SKIPPED]
↓ Create tsconfig.json [SKIPPED]
↓ Create eslint config [SKIPPED]
✔ Add package scripts
✖ /private/var/folders/9n/72xhx63x5gzcqchlbd5skvz40003jl/T/reheasal-smoke_-27421-IDaiBHi2uUMH/urn-utils/tsc…
◼ Convert JS files to TS
error:   SyntaxError: /private/var/folders/9n/72xhx63x5gzcqchlbd5skvz40003jl/T/reheasal-smoke_-27421-IDaiBHi2uUMH/urn-utils/tsconfig.json: Unexpected token / in JSON at position 161
```

lynchbomb:
Ok we need to use json5 for parsing then

lynchbomb:
Fixed with #883

PR #883, opened 2023-03-24
fix: refactor to use json5.parse for read of tsconfig

wondersloth:
In parts of CLI package we were using `readJSONSync` from `fs` for reading `tsconfig.json`. This doesn't work for `.json` files that have comments, or unquoted keys.

See #876 for example where an internal `tsconfig.json` from the jeeves task created config with a comment.

### Added
- **packages/util/** Add `readTSConfig` method using `json5.parse()`
### Changed
- **packages/util** Refactored `writeTSConfig()` to allow for passing of a config data. Expose defaults at DEFAULT_TS_CONFIG.
- **packages/cli** Refactor usages where reading and writing to `tsconfig.json` to use  `readTSConfig` and `writeTSConfig` from `@rehearsal/util`.
- **packages/cli** Refactor tests in `analyze.test.ts` to setup and tear down projects more succinctly using the fixturify API.

### Removed
- Usage of `json5` in **packages/cli** as it's no longer used.

Notes: 
- We should not not drop usage `get-tsconfig` as it it collapses and resolved the tsconfig down to it's final shape. This is critical in it's use case.
- We can't use `get-tsconfig` everywhere because it expands all globs patterns for the tsconfig.include property, which is not what we want in most cases.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -55,7 +55,6 @@
     "fs-extra": "^11.1.0",
     "get-tsconfig": "^4.4.0",
     "js-yaml": "^4.1.0",
-    "json5": "2.2.3",
     "listr2": "^5.0.7",
     "minimatch": "^7.4.2",
     "semver": "^7.3.8",
--- a/packages/cli/src/commands/migrate/tasks/analyze.ts
+++ b/packages/cli/src/commands/migrate/tasks/analyze.ts
@@ -5,9 +5,8 @@ import {
   getMigrationStrategy,
   SourceFile,
 } from '@rehearsal/migration-graph';
-import { determineProjectName } from '@rehearsal/utils';
+import { determineProjectName, readTSConfig, writeTSConfig } from '@rehearsal/utils';
 import debug from 'debug';
-import { readJSONSync, writeJSONSync } from 'fs-extra/esm';
 import { ListrDefaultRenderer, ListrTask } from 'listr2';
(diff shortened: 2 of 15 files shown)


packages/cli/test/commands/upgrade.test.ts
```diff
@@ -113,7 +112,7 @@ describe('upgrade:command typescript@next', () => {
     expect(result.stdout).contain(`Rehearsing with typescript@${latestPublishedTSVersion}`);
     expect(existsSync(reportFile)).toBeTruthy();
 
-    const report = readJSONSync(reportFile) as import('@rehearsal/reporter').Report;
+    const report = readJSON(reportFile) as import('@rehearsal/reporter').Report;
```
wondersloth, line 115: This could be rewritten as:
```suggestion
    const report = readJSON< import('@rehearsal/reporter').Report>(reportFile);
```

Though this is a test and it's minor.

wondersloth:
Overall I think using JSON5 is bad because it doesn't quote attributes when writing.

There is a package `get-tsconfig` that is being used to readTsConfig, that we should use consistently to close out this issue.

- [x] Refactor reading of tsconfig to use a singular method readTsConfig() from utils that will use `get-tsconfig` package. This will 
- [x] Drop refactor to use a readJSON and writeJSON method with JSON5.

lynchbomb (approved):

PR #885, opened 2023-03-25
Update test-and-lint.yml

chadhietala:
I noticed that #879 removed the unit tests from CI but it should actually run. This also pulls apart the steps for lint, test, slow-test so they can run in parallel.
Screenshot 2023-03-25 at 12 49 18 PM


lynchbomb:
test:slow is additive to test. ie test:slow should be all that is needed (shouldn’t need test:slow plus test)

PR #886, opened 2023-03-26
fix(cli): generator for interactive mode

### LINKED ISSUE #831: migrate is overly verbose on `processing file:`
See attached video. The `git mv` is working correctly with interactive mode - however `processing file:` is now logging every `.ts` file migrated and the remaining `.js` files in the migrated package. The desired logging should only print `processing file:` once.

https://user-images.githubusercontent.com/10506014/223883334-c413b02c-cb83-4d23-81a2-0dbee32088f2.mov

haochuan:
Fixes #831 

##  In `PluginsRunner`

- Change `*run` to a generator, which will yield at each file.

##  In `migrate` package

- Use `*run`.

## In `regen` package, and `upgrade` CLI command

Use this to call the generator:

```
for await (const _ of runner.run(fileNames)) {
        // no ops for yield
      }
```

## Re-structure the loop and conditions in CLI `convert` task

- If we have files to migrate, continue
- Iteration `*run`
  - In ci mode, just tracking files in every iteration
  - In interactive mode, stop at each `yield` and prompt for options


packages/service/src/plugin.ts
```diff
@@ -41,7 +41,7 @@ export class PluginsRunner {
     return this;
   }
 
-  async run(fileNames: string[], logger?: PluginLogger): Promise<void> {
```
lynchbomb, line 44: What is the use-case for keeping run? Why not have the generator as the default and when we don’t need to process the yielded file just drain the entire process.

haochuan: @lynchbomb 

Two reasons to keep `run` while writing these:

1. I though the `setImmediate(next)` in the old `run` would still be a little little faster than just yield, when we just want to run all of them at once.
2. `await runAll(fileNames)` looks a little cleaner than `for await (const _ of run(fileName)) { // do thing }`

But I'm totally open to just keeping the generator version, since the difference is just too little.

haochuan: @lynchbomb let me know. If you think it's better to just keep the generator version `run`, I'm happy to change to that.

egorio: I, the opposite, would keep only `run` with Promise output with a callback and hide the generator inside, like this (example with migrate):

Instead of
```
for await (const f of migrate(input)) {
   migratedFiles.push(f);
}
```
do 
```
await migrate(input, (file) => migratedFiles.push(file))
```

If we want to keep the generator as output - I think it's better to keep only `run => AsyncGenerator` to avoid confusion... but I don't like the idea that devs need to handle yields just by running plugins (`for await (const _ of run(fileName)) { // do thing }`).

egorio (approved):

PR #889, opened 2023-03-27
feat(cli): package option

### LINKED ISSUE #576: Package selections in non-interactive mode
Currently in interactive mode we are doing two things:
1. Let user select child package to run partial migration (if there is any)
2. Go through migrated files one by one and ask for actions (accept/edit/discard)

I feel it makes more sense to also enable No.1 in non-interactive mode, and No.2 is the only difference between interactive and normal mode.

@chadhietala @lynchbomb Any idea/thought on this one?

haochuan:
Fixes #576

- Add new option `-p, --package` for pass package to the CLI
  - In CI mode, now we can run migrate on a specific package
  - In Non CI mode, pass a valid package will skip the "package selection" prompt
- Rename `-p --basePath` to `-b --basePath` to remove the conflict, we don't use and expose `--basePath` now so it should be good.

### DIFF
--- a/packages/cli/src/commands/migrate/index.ts
+++ b/packages/cli/src/commands/migrate/index.ts
@@ -30,7 +30,7 @@ migrateCommand
   // always use default process.cwd()
   .addCommand(initCommand)
   .addOption(
-    new Option('-p, --basePath <project base path>', 'base directory of your project')
+    new Option('-b, --basePath <project base path>', 'base directory of your project')
       .default(process.cwd())
       // use argParser to ensure process.cwd() is basePath
       // even use passes anything accidentally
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/migrate/index.ts
```diff
@@ -47,6 +47,11 @@ migrateCommand
     `path to a entrypoint file inside your project(${process.cwd()})`,
     ''
   )
+  .option(
+    '-p, --package <relative path to target package>',
+    `run migrate againt a specific package in your project(${process.cwd()}) `,
```
lynchbomb, line 52: `run migrate against a specific child-package in your project`

lynchbomb (changes requested):

lynchbomb (approved):

PR #892, opened 2023-03-27
fix(service): move test-support to devDep

lynchbomb:
This PR fixes an incorrect import of @rehearsal/test-support as a bundled dependency of @rehearsal/service. This should be a dev-dependency only.


wondersloth (approved):

lynchbomb:
The smoke test is going to fail until this is merged

PR #897, opened 2023-03-28
fix: run builds topologically

cafreeman:


chadhietala:
But it does via [project references](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/tsconfig.json#L7).

egorio (approved):

lynchbomb:
I dont think we need this. @cafreeman is going to confirm when he adds the missing paths to the package tsconfig his issue is resolved. DO NOT MERGE.

PR #903, opened 2023-03-29
fix: unable to get bin path with npm@9

haochuan:
We cannot rely on using `<package manager> bin` to get the bin dir for executables, since `npm>=9` already dropped the subcommand `bin`.

Assuming there is a executable `foo`:
I think it still better to try `<package manager> bin foo` first, then fallback to check if `foo` exists in `node_modules/.bin/foo` and return `node_modules/.bin` as the dir.

It is complicated to test in different scenarios, since we need to create "fake" `npm/yarn/pnpm` with different versions. There would be a high risk to pollute other tests and cause conflict so I just added a TODO in the test.


lynchbomb:
@haochuan I think you need to pull in @brendaHuang changes; which should make your tests pass

PR #913, opened 2023-03-31
feat: enable prettier plugin as the final step

egorio:
Rehearsal will conditionally choose to use Prettier or ESlint for formatting based on the availability of a prettier config file.
If the target project has a prettier config - then prettier will be used for formatting files, if no file exists - eslint will be used. 

I found having prettier and eslint plugins separated is hard to support due to additional logic we need to carry to choose which plugin to use for a particular file. I'm planning to reorg plugins to `lint-report.plugin` and `formatter.plugin` (including eslint and prettier + additional logic to choose which one to use) the next PR.


packages/migrate/test/fixtures/project/.prettierrc
```diff
@@ -0,0 +1,20 @@
+{
```
lynchbomb, line 1: Since we are shifting this work into prettier. We need to be sure the migrating project in question has a similar config.

packages/plugins/src/plugins/prettier.plugin.ts
```diff
@@ -1,25 +1,22 @@
 import { Plugin, PluginOptions, type PluginResult, PluginsRunnerContext } from '@rehearsal/service';
-import { format, type Options } from 'prettier';
+import pretiter from 'prettier';
```
lynchbomb, line 2: typo `prettier`

packages/plugins/src/plugins/prettier.plugin.ts
```diff
@@ -1,25 +1,22 @@
...
-    context: PluginsRunnerContext,
-    options: PrettierPluginOptions
-  ): PluginResult {
+export class PrettierPlugin implements Plugin<PluginOptions> {
+  async run(fileName: string, context: PluginsRunnerContext): PluginResult {
     const text = context.service.getFileText(fileName);
 
     try {
-      const result = format(text, options);
+      const prettierOptions = pretiter.resolveConfig.sync(fileName) || {};
```
lynchbomb, line 16: typo `prettier`

packages/plugins/test/prettier.plugin.test.ts
```diff
@@ -4,9 +4,9 @@ import { PrettierPlugin } from '../src/index.js';
 import { initProject, mockPluginRunnerContext } from './utils.js';
 
 describe('Test PrettierPlugin', () => {
-  test('run', async () => {
-    const project = await initProject('prettier-test', {
-      'index.ts': './test/fixtures/prettier.ts.fixture',
+  test('rung default', async () => {
```
lynchbomb, line 7: `run`

lynchbomb (changes requested):
Looks good. Couple typos. Also we need to validate the project being migrated has a prettier config, as we are now relying on this for formatting.

lynchbomb (approved):

PR #914, opened 2023-03-31
test: add config-ts ember tests

brendaHuang-2023:
This PR is the most important part of the original PR #870. This PR is created for readability.

It adds ember tests for the `config-ts` task. The iterations look small, but each of them is a full iteration of the `config-ts` task under different conditions.

### DIFF
--- a/packages/cli/test/commands/migrate/config-ts.ember-app.test.ts
+++ b/packages/cli/test/commands/migrate/config-ts.ember-app.test.ts
@@ -32,11 +32,12 @@ describe('Task: config-ts ember app', () => {
         output = '';
 
         vi.spyOn(console, 'info').mockImplementation((chunk: string) => {
-          chunk = chunk.replace(new RegExp(project.baseDir, 'g'), '<tmp-path>');
+          chunk = chunk.replace(new RegExp(`${project.baseDir}`, 'g'), '<tmp-path>');
           output += `${chunk}\n`;
         });
 
-        vi.spyOn(console, 'log').mockImplementation((chunk) => {
+        vi.spyOn(console, 'log').mockImplementation((chunk: string) => {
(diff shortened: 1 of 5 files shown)


packages/cli/test/commands/migrate/config-ts.ember-app.test.ts
```diff
@@ -0,0 +1,165 @@
...
+        // Prepare old tsconfig
+        const oldTsConfig = { compilerOptions: { strict: false } };
+        writeJSONSync(resolve(project.baseDir, 'tsconfig.json'), oldTsConfig);
+
+        await runTsConfig(project.baseDir);
+
+        const tsConfig = readJSONSync(resolve(project.baseDir, 'tsconfig.json')) as TSConfig;
+
+        expect(tsConfig.compilerOptions.strict).toBeTruthy();
+        // Do not use snapshot here since there is absolute path in output
```
lynchbomb, line 71: see my previous comment on dynamic paths. you'll need to sanitize the path from the input like we are doing in tests so the snapshot contains eg https://github.com/rehearsal-js/rehearsal-js/blob/f89124d5050bbf787e1524602f1ab97831ff51f6/packages/cli/test/commands/migrate/__snapshots__/convert.test.ts.snap#L40

lynchbomb: ... same for all comments in this PR around `// Do not use snapshot here since there is absolute path in output`

lynchbomb (changes requested):

packages/cli/test/commands/migrate/config-ts.ember-app.test.ts
brendaHuang-2023: fixed

lynchbomb (approved):

PR #915, opened 2023-03-31
feat: add glint plugin fixes to reporter

cafreeman:
Hooks glint plugins into the Rehearsal reporting system.

### DIFF
--- a/packages/plugins/src/plugins/glint-report.plugin.ts
+++ b/packages/plugins/src/plugins/glint-report.plugin.ts
@@ -1,5 +1,5 @@
 import { pathUtils } from '@glint/core';
-import { DiagnosticWithContext, hints } from '@rehearsal/codefixes';
+import { DiagnosticWithContext, hints, getDiagnosticOrder } from '@rehearsal/codefixes';
 import { Reporter } from '@rehearsal/reporter';
 import {
   GlintService,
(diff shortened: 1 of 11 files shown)


packages/plugins/src/plugins/glint-report.plugin.ts
```diff
@@ -37,10 +41,32 @@ export class GlintReportPlugin implements Plugin<PluginOptions> {
     return Promise.resolve([fileName]);
   }
 
+  getDiagnostics(service: Service, fileName: string): DiagnosticWithContext[] {
+    const languageService = service.getLanguageService();
+    const program = languageService.getProgram()!;
+    const checker = program.getTypeChecker();
+
+    const diagnostics = service.getDiagnostics(fileName);
```
chadhietala, line 49: I think you want to go through this https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/codefixes/src/get-diagnostics.ts#L26

cafreeman: Good call, fixed.

egorio (dismissed):

lynchbomb:
@cafreeman we should validate the glint comments don't cause issue with the various @rehearsal/reporter formatters eg. SARIF, sonar

For this PR could you expand the tests in @rehearsal/reporter to include examples with glint comments in the report

lynchbomb:
otherwise LGTM

cafreeman:
> @cafreeman we should validate the glint comments don't cause issue with the various @rehearsal/reporter formatters eg. SARIF, sonar
> 
> For this PR could you expand the tests in @rehearsal/reporter to include examples with glint comments in the report

Discussed in Slack. Since the glint comments actually show up as regular TS errors as far as the report is concerned, nothing really changes in the formatters here so there isn't anything to test

lynchbomb (approved):

PR #943, opened 2023-04-07
fix: add resolution for @swc/core@1.3.44

wondersloth:
Adding a resolution for `@swc/core@1.3.44` because of this packaging bug:
https://github.com/swc-project/swc/issues/7203

Made the resolution in `migration-graph-ember` because that's where it's being used and the install of the cli fails when this package installs when `1.3.46` is used when we invoke `volta install @rehearsal/cli` with our Jeeves migration script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,9 +97,6 @@
     "vitest": "^0.29.2",
     "walk-sync": "^3.0.0"
   },
-  "resolutions": {
-    "@swc/core": "1.3.44"
-  },
   "packageManager": "pnpm@7.27.0",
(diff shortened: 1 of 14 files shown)


wondersloth:
I think the resolution should be in package.json for `migration-graph-ember`
https://www.npmjs.com/package/@rehearsal/migration-graph-ember?activeTab=code

Thinking of how the dependencies are installed the root project doesn't affect this published file.

egorio (approved):

lynchbomb (approved):

PR #965, opened 2023-04-11
chore(all): node env and pnpm to latest LTS

lynchbomb:
This PR
- bumps Node to latest LTS (18.15.0)
- pnpm latest (8.2.0)
- regenerates lock file
- bumps both in all gh-actions

### DIFF
--- a/.github/workflows/get-diagnostics.yml
+++ b/.github/workflows/get-diagnostics.yml
@@ -17,10 +17,10 @@ jobs:
           ref: ${{ github.head_ref }}
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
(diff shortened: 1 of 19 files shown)


.github/workflows/get-diagnostics.yml
```diff
@@ -17,10 +17,10 @@ jobs:
           ref: ${{ github.head_ref }}
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: 18
```
chadhietala, line 20: you might want to use the `matrix` feature here and test 16. It's going to be a while until we can use node 18 at LI.

egorio (approved):

PR #969, opened 2023-04-12
fix: fix graph sorting

chadhietala:
Prior to this PR we were not collecting any of the dependencies of packages. The callstack is:

```
EmberAppProjectGraph.discovery
  EmberAppProjectGraph.addPackageToGraph
    ProjectGraph.addPackageToGraph
      ProjectGraph.discoverEdgesFromDependencies
        ProjectGraph.findInternalPackageDependencies
```

Inside of `ProjectGraph.findInternalPackageDependencies` we [filter dependencies](https://github.com/rehearsal-js/rehearsal-js/blob/a3be6fe1cfc50881d288db1abc1edc6cda7d672c/packages/migration-graph-shared/src/entities/project-graph.ts#L133) to only the things we have seen by checking `this.discoveredPackages`. `this.discoveredPackages` is being materialized as we insert packages and will never properly attach edges without this.

In addition, I have renamed this `topSort` since the algo is not actually topsort.


packages/migration-graph-ember/test/utils/discover-ember-packages.test.ts
```diff
@@ -1,8 +1,37 @@
...
+    projectGraph.discover();
+    const nodes = projectGraph.graph.topSort();
+
+    expect(nodes.map((n) => n.content.pkg?.packageName)).toEqual([
+      '@company/b',
+      '@company/a',
+      '@company/c',
+      '@company/d',
+      'monorepo',
+    ]);
```
wondersloth: This should be added to [`EmberAppProjectGraph-test.ts`](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migration-graph-ember/test/entities/ember-app-project-graph.test.ts)

As it's largely asserting that the order was correct given a project fixture.

This test could stay here if we're calling discoverEmberPackages explicitly and asserting it's output from the method.

wondersloth (dismissed):

packages/test-support/src/project.ts
```diff
@@ -200,8 +204,6 @@ export function getEmber4AppProject(): Project {
 }
 
 export async function setupProject(project: Project): Promise<Project> {
-  const { name: tmpDir } = tmp.dirSync();
-  project.baseDir = tmpDir;
```
chadhietala, line 204: You don't need to create your own tmp dir. Fixturify-project generates one for you when you `write()`

lynchbomb (approved):

Issue #964, opened 2023-04-11
Incorrect Resolution For Addons


chadhietala:
This [line](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/migrate/src/glint-utils.ts#L23) assumes there is a node_modules at the `basePath` however this is not always the case. I believe the idea is that `basePath` is always supposed to be `process.cwd()` however if you follow the code back you'll see that `targetPackagePath` is [being using](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/migrate/tasks/convert.ts#L57). When I tried to fix this tons of things broke.

egorio:
I also was confused about `basePath` / `cwd` / `targetPackagePath` / `--entrypoint` / `--package` paths usage when working on [this issue ](https://github.com/rehearsal-js/rehearsal-js/pull/939/files).

It's where we expect to have package.json / node_modules / what is the root path for relative paths etc.
I think we need to define few path wich better names at start and later in the code.

PR #971, opened 2023-04-12
fix: correctly resolve addon paths

cafreeman:
Closes https://github.com/rehearsal-js/rehearsal-js/issues/964

Uses https://www.npmjs.com/package/resolve-package-path to correctly resolve package paths even in cases where the `basePath` is not the same as `process.cwd` or the root of the project (e.g. when running something like `rehearsal migrate --package ./packages/foo` in a monorepo)


chadhietala (approved):
Seems correct. Lets see what the tests say.

PR #977, opened 2023-04-13
internal: use magicstring for ignore comments

chadhietala:
For a long time now we have relied on `while (diagnostics.length) {...}` as the mechanism to ensure that modifications to files were inserted correctly because the lines and columns would change as we make the modifications. While we typically would get these insertions correct, it meant that we had to effectively run the typechecker on every turn of the loop to ensure the line and cols of the diagnostics were correct. The unintended side effect of this was that different diagnostic codes could enter the `diagnostics` array on a partially migrated file. In addition, when there were bugs we would get stuck in an infinite loop and have no real recurse on how to reproduce the bug locally. In addition to these issues there were cases where we were not accounting for comment insertion correctly, which lead us to using multiline comment syntax for `@ts-expect-error` or inserting the comments above statements when they should have been stuck closer to expressions.

This PR does the following things:

- Directly uses the same logic as ts-migrate for inserting `ts-expect-error` comments, which fixes cases we did not handle correctly but also tracks which lines have been modified so we don't find our selves modify lines that have already been modified. This will drastically increase reproducibility.
- Moves to a single pass loop for apply the comments
- Uses [magic-string](https://github.com/Rich-Harris/magic-string/) for change tracking modifications. The API assumes that all the insertions are relative to the original document offsets, which is exactly what we need.
- Separates out adding comments from reporting to always follow this flow `transform -> format -> comment -> format -> report`
- Reporting is now non-destructive in that it removes all of the comments but retains the whitespace where comments normally are. This gives us an accurate line and col position for the report without [having to guess](https://github.com/rehearsal-js/rehearsal-js/pull/977/files#diff-661e854c962a54714baad2261ba5c5ab47b5c518ced31145f6898e7e9e977ca6L5) which line the comment got inserted on.

*Next Steps*

- Do a similar single pass loop to diagnostic.fix plugins
- Change the execution of files from a single file moving through each stage of the. queue, to all files completing each stage of the queue before, going to the next stage
- Regen seems to know nothing about Glint so that needs to be integrated


packages/plugins/src/plugins/diagnostic-comment.plugin.ts
```diff
@@ -0,0 +1,238 @@
...
+  getConditionalCommentPos,
+  inJsxText,
+  inTemplateExpressionText,
+  onMultilineConditionalTokenLine,
+} from './utils.js';
+import type MS from 'magic-string';
+
+const require = createRequire(import.meta.url);
+// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+const MagicString = require('magic-string');
```
lynchbomb, line 24: why the require? magic-string exports as ESM `import MagicString from 'magic-string';`

packages/plugins/src/plugins/glint-comment.plugin.ts
```diff
@@ -0,0 +1,219 @@
...
+import debug from 'debug';
+import { getLocation } from '../helpers.js';
+import { getConditionalCommentPos, onMultilineConditionalTokenLine } from './utils.js';
+import type MS from 'magic-string';
+
+const DEBUG_CALLBACK = debug('rehearsal:plugins:glint-comment');
+
+const require = createRequire(import.meta.url);
+// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+const MagicString = require('magic-string');
```
lynchbomb, line 24: same `import MagicString from 'magic-string';`

lynchbomb (dismissed):
Minor to swap the import of magic-string to ESM

packages/plugins/src/plugins/diagnostic-comment.plugin.ts
chadhietala: I have tried all combinations of this and it just doesn't work. When you do `import MagicString from 'magic-string';` and you try and call it you get the following type errors.

Screenshot 2023-04-14 at 10 24 19 AM

When you invoke it as `new MagicString.default` you don't get type errors.

Screenshot 2023-04-14 at 10 24 33 AM

However you get runtime errors

Screenshot 2023-04-14 at 10 24 41 AM

lynchbomb (dismissed):

egorio:
The `-report` behavior is not 100% correct....

egorio (dismissed):

PR #978, opened 2023-04-13
fix(cli): include the actual error when throwing dep install

lynchbomb:
This PR:

- include the actual error when dep install fails rather than only the abstract error


cafreeman (dismissed):

chadhietala:
@lynchbomb I think this can be closed given the pivot.

lynchbomb:
Pivot

Issue #975, opened 2023-04-13
Bug: migrate fails when a project doesn't have `@glint/core` installed


wondersloth:
Internally

```
$ <checkout> fedex-realtime-docs
$ cd fedex-realtime-docs
$ yarn install
$ volta install @linkedin/rehearsal-cli-config
$ rehearsal-cli-config migrate-config init --ci
$ yarn add --dev @rehearsal/cli
$ yarn rehearsal migrate --ci
```

```
✖ Cannot find package '@glint/core' imported from /Users/medwards/Code/muuto/test-packages/urn-utils/node_modules/@rehearsal/ser…
error:   Cannot find package '@glint/core' imported from /Users/medwards/Code/muuto/test-packages/urn-utils/node_modules/@rehearsal/service/dist/src/glint-service.js
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@glint/core' imported from /Users/medwards/Code/muuto/test-packages/urn-utils/node_modules/@rehearsal/service/dist/src/glint-service.js
    at new NodeError (node:internal/errors:372:5)
    at packageResolve (node:internal/modules/esm/resolve:954:9)
    at moduleResolve (node:internal/modules/esm/resolve:1003:20)
    at defaultResolve (node:internal/modules/esm/resolve:1218:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:580:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:294:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:80:40)
    at link (node:internal/modules/esm/module_job:78:36)
```

Migrate fails because it imports `glint-service`
from `@rehaersal/service`  and `@glint/core` is a `peerDependency` in service.
We need to conditionally import [GlintService](https://github.com/rehearsal-js/rehearsal-js/blob/c172d4fda4979d7b23744d6238708b638d0ebba0/packages/migrate/src/migrate.ts#L111)
possibly just refactor to a factory where we lazy import.

PR #983, opened 2023-04-13
fix: use dynamic imports for glint-related modules

cafreeman:
Hopfully closes https://github.com/rehearsal-js/rehearsal-js/issues/975 by making all of the glint-related modules dynamic imports so that we don't even attempt to import things like `GlintService` when we're not in an ember project

### DIFF
--- a/packages/migrate/src/glint-utils.ts
+++ b/packages/migrate/src/glint-utils.ts
@@ -4,6 +4,8 @@ import { readPackageJson } from '@rehearsal/migration-graph-shared';
 import { requirePackageMain } from '@rehearsal/migration-graph-ember';
 import resolvePackagePath from 'resolve-package-path';
 import type { PackageJson, TsConfigJson } from 'type-fest';
+import type { GlintService } from '@rehearsal/service';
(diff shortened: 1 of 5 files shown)


packages/migrate/src/glint-utils.ts
```diff
@@ -87,3 +89,30 @@ export async function shouldUseGlint(basePath: string): Promise<boolean> {
     return GLINT_PROJECT_FILES.includes(pkgName);
   });
 }
+
+export const DummyPlugin = {
```
wondersloth, line 93: This should implement Plugin

https://github.com/rehearsal-js/rehearsal-js/blob/c67e693979aa6becabfacbd65f64dafa8b1fc884/packages/service/src/plugin.ts#L5

wondersloth (changes requested):
Minor suggestion, otherwise looks good.

packages/migrate/src/glint-utils.ts
cafreeman: It does now by virtue of how the TS type system works (Plugin is a thing with a run method that returns `Promise<Array>` so the type system already sees it as a Plugin

wondersloth (approved):

PR #972, opened 2023-04-12
chore(all): bump deps and remove unused deps, coverage

lynchbomb:
This PR:

- adds coverage to "fast" tests 
- bumps deps/devDeps
- removes unused deps
- updated @typescript-eslint/no-unused-vars to allow for "_"

PR #986, opened 2023-04-14
fix: update lockfile

cafreeman:
Follow-on to https://github.com/rehearsal-js/rehearsal-js/pull/972 that commits the required lockfile changes


cafreeman:
Closing because we need to actually get everyone on the same pnpm version first.

PR #987, opened 2023-04-14
internal: use magic-string for fixing

chadhietala:
This is a follow on to #977 to move code modification to use magic-string for change tracking. Like #977, we are moving to a more conservative model to limit the amount of modifications we do, however instead of keeping track of lines we edit we track offsets relative to the original file. This obviously limits the amount of errors we can fix, however we need to draw a baseline of conservative but reproducible outcomes. We can expand the types we allow through at the same offset in the future.

I noticed that we were relying on re-checking the file of each turn of the loop to resolve parameter types in `addMissingTypesBasedOnInheritance`, however we can [continue walking up](https://github.com/rehearsal-js/rehearsal-js/pull/987/files#diff-17e38cc310e673b5b57aae2003b38325b0bfba76cc2dac216bffc4d8c0a7a629R139) the class hierarchy to find static types.

This PR does the following:

- Moves fix plugins to use magic-string to do change tracking
- Tracks the offsets of where we placed new types
- Fix plugins now do only 2 passes, 1 pass to fix errors and 1 pass to apply any suggestions
- Changes `addMissingTypesBasedOnInheritance` to walk the class hierarchy when we get loose types

# Follow Ups

- #988 is a cleanup of the plugin queue to create a new instance of the plugin on a per-file bases so we can stop doing [state reseting](https://github.com/rehearsal-js/rehearsal-js/pull/987/files#diff-11e3e65b5b4fefbcc62fbc0826e0036b9a8755b79ad29a33ff8149b126a17152R52) in the plugins

### DIFF
--- a/packages/codefixes/src/fixes/addMissingTypesBasedOnInheritance.ts
+++ b/packages/codefixes/src/fixes/addMissingTypesBasedOnInheritance.ts
@@ -136,6 +136,17 @@ export class AddMissingTypesBasedOnInheritanceCodeFix implements CodeFix {
 
         // Skipping "non-strict types"
         if (!type || type === 'any' || type === 'object') {
+          // If the type was too loose but we have an accesortor of the current class
+          // recurse with the parent
+          const parentClassMethod = findAncestor(
+            ancestorMethodSymbol.declarations?.[0],
+            isMethodDeclaration
+          );
+
+          if (parentClassMethod) {
+            return this.findInheritedType(parentClassMethod, checker, finder);
+          }
+
           continue;
         }
 
--- a/packages/codefixes/src/get-diagnostics.ts
+++ b/packages/codefixes/src/get-diagnostics.ts
@@ -16,6 +16,7 @@ const PRIORITIZED_CODES: number[] = [
   Diagnostics.TS7046.code,
   Diagnostics.TS7050.code,
   Diagnostics.TS2612.code,
+  Diagnostics.TS4073.code,
 ];
 
 /**
--- a/packages/codefixes/test/__snapshots__/base-codefixes.test.ts.snap
+++ b/packages/codefixes/test/__snapshots__/base-codefixes.test.ts.snap
@@ -111,7 +111,8 @@ export default <P extends InjectedProps>(Component: React.ComponentType<P>) =>
         [...prev.todos, todo];
       });
     };
-    props: P | undefined;
+    // @ts-expect-error @rehearsal TODO TS2564: Property 'props' has no initializer and is not definitely assigned in the constructor.
+    props: P;
 
     render() {
       return (
--- a/packages/migrate/test/migrate.test.ts
+++ b/packages/migrate/test/migrate.test.ts
@@ -373,9 +373,9 @@ export default class Hello extends Component {
 import { inject as service } from "@ember/service";
 
 export default class Salutation extends Component {
-  @service locale: { current: () => string } | undefined;
+  // @ts-expect-error @rehearsal TODO TS2564: Property 'locale' has no initializer and is not definitely assigned in the constructor.
+  @service locale: { current: () => string };
   get name() {
-    // @ts-expect-error @rehearsal TODO TS2532: Object is possibly 'undefined'.
     if (this.locale.current() == "en-US") {
       return "Bob";
     }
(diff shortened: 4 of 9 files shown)


lynchbomb (approved):

egorio:
These changes completely broke the original idea and behavior of `diagnostic.fix` plugin - modifying files until all errors are fixed or commented (skipped).

The original behavior was:
```
let diagnostics = getDiagnostics()
while (diagnostics) {
  fix()
  diagnostics = getDiagnostics() // get updated diagnostics
}
```

the new one:
```
let diagnostics = getDiagnostics()
while (diagnostics) {
  fix()
}
```

Originally we had a priority order of codefixes (imports, types from jsdoc) and applied them first. This helped to clean some errors in the files available originally... because we always updated the diagnostics list after applying them.

In the new approach, we don't update the diagnostics list after applying codefix, so we are trying to fix errors down the list which possibly already been fixed, for example:

- two diagnostics added the same `import` statement - then 2 same import statements will be added to the file.
- two codefixes can be applied at the same node (for example types from inference and types from jsdoc) - type will be applied twice
- one type relies on another which can be inferred (return rely on input value) - only the known type will be inferred, but all related won't

chadhietala:
> This obviously limits the amount of errors we can fix, however we need to draw a baseline of conservative but reproducible outcomes.

This is expected.

lynchbomb:
Most likely we will land somewhere between the two implementations. Once we iterate and settle (some future state). The recursive function was error prone and very difficult (if not impossible) to repeatably reproduce. 

We obviously do not want to introduce runtime errors and should always lean away from this.

lynchbomb:
Forgive the simplicity but have we tried just adding a counter outside of the loop and exiting after eg 2-3 turns? We know 1 turn can produce runtime errors and infinite turns isn’t repeatable etc.

egorio:
@chadhietala @lynchbomb 

> The recursive function was error-prone and very difficult (if not impossible) to repeatably reproduce.
> Adding a counter outside of the loop and exiting after eg 2-3 turns

We didn't have a recursive function (as it appears to be)... we had a loop over **first-time available** diagnostics + only **new**. We tried t fix every error only once and [skipped it next attempts](https://github.com/rehearsal-js/rehearsal-js/blob/v1.0.20-beta/packages/plugins/src/plugins/diagnostic-fix.plugin.ts#L168C11-L178) (after getting new diagnostics).
 
We didn't need a counter for the same reason... we can't have an infinitive number of errors, and the loop stops once it tried to fix each available error only once.

**Original workflow:**
```
Cycle 1:
We have errors on lines 1, 2, 3, 4, 5
- Trying to fix 1. 
- Fixed... let's say this also fixed 3, 4, but bring errors on lines 6, 7
- Get new errors = 2, 5, 6, 7

Cycle 2:
- Now we have errors on lines 2, 5, 6, 7
- Trying to fix 2. 
- Can't fix... added it to `attemptedToFix` = [2]
- Get new errors = 2, 5, 6, 7... still the same

Cycle 3:
- Now we have errors on lines 2, 5, 6, 7
- Skiped error on line 2 (because it's already in attemptedToFix)
- Trying to fix 5. 
- Fixed
- Get new errors = 2, 6, 7

Cycle 4:
- We have errors on lines 2, 6, 7
- Skiped error on line 2 (because it's already in attemptedToFix)
- Fixed... this also fixed error 7, but added 8, 9 
- Get new errors = 2, 8, 9 

Cycle 5:
- Now we have errors on lines 2, 8, 9
- Skiped error on line 2 (because it's already in attemptedToFix)
- Trying to fix 8. 
- Can't fix... added it to `attemptedToFix` = [2,8]
- Get new errors = 2, 8, 9... still the same

Cycle 6:
- Now we have errors on lines 2, 8, 9
- Skiped error on lines 2 and 8 (because it's already in attemptedToFix)
- Trying to fix 9. 
- Can't fix... added it to `attemptedToFix` = [2,8,9]
- Get new errors = 2, 8, 9... still the same

Cycle 7:
- Now we have errors on lines 2, 8, 9
- Skiped error on lines 2, 8, 9 (because it's already in attemptedToFix)

End
```

So the [while](https://github.com/rehearsal-js/rehearsal-js/blob/v1.0.20-beta/packages/plugins/src/plugins/diagnostic-fix.plugin.ts#L50) loop was max out to **initial errors + new errors only** cycles.

Issue #585, opened 2023-01-09
Skip plugins if there is nothing to fix on file


egorio:
After https://github.com/rehearsal-js/rehearsal-js/issues/584

The idea is to create a plugin to test if there are any diagnostic messages on the file and run the rest of plugins only if there are any.

The way to do it is:

create a queue of files to apply plugins.
create a plugin to check for errors in the file.
allow the plugin to create a queue of files for the next plugin (means if there are no diagnostic messages for the file, it won't be added to queue for linter plugin.

Issue #584, opened 2023-01-09
Refactor rehearsal workflow to "for plugins - for files"


egorio:
Rehearsal (migrate) current workflow is:

for (files) { for (plugins) { … } }
let’s assume we have 1, 2, 3, 4 files.
we apply all plugins to each file one by one... the first one is 1.
to fix the issue in the 3 rd we need to modify 1… but we already applied some plugins to it, means the modification of this file, at least, can brake its formatting

So we probably need to do the instead (like `ts-migrate` does):

`for (plugins) { for (files) { … } }`

PR #988, opened 2023-04-16
internal: restructure plugins

chadhietala:
This PR restructures the plugin execution loop and provides an abstract base class that plugins can inherit from. Prior to these changes we had to [reset the class state](https://github.com/rehearsal-js/rehearsal-js/pull/988/files#diff-11e3e65b5b4fefbcc62fbc0826e0036b9a8755b79ad29a33ff8149b126a17152L205) on a per file basis. Instead of registering instances into the queue we can register the constructors and let the queue executor be responsible for constructing them with the correct options. 

This PR does not change any user facing behavior and only cleans up how we author, execute and register plugins.

### DIFF
--- a/packages/migrate/src/glint-utils.ts
+++ b/packages/migrate/src/glint-utils.ts
@@ -3,9 +3,13 @@ import path from 'node:path';
 import { readPackageJson } from '@rehearsal/migration-graph-shared';
 import { requirePackageMain } from '@rehearsal/migration-graph-ember';
 import resolvePackagePath from 'resolve-package-path';
+import type { GlintFixPlugin, GlintCommentPlugin, GlintReportPlugin } from '@rehearsal/plugins';
 import type { PackageJson, TsConfigJson } from 'type-fest';
 import type { GlintService } from '@rehearsal/service';
-import type { GlintCommentPlugin, GlintFixPlugin, GlintReportPlugin } from '@rehearsal/plugins';
+
+type GlintFixPluginCtor = typeof GlintFixPlugin;
+type GlintReportPluginCtor = typeof GlintReportPlugin;
+type GlintCommentPluginCtor = typeof GlintCommentPlugin;
 
 // The list of extensions that we expect to be handled by Glint{Fix,Check} plugins. Note that
 // in any ember/glimmer project, we'll use the glint *service* for all files. This list is only
@@ -90,12 +94,6 @@ export async function shouldUseGlint(basePath: string): Promise<boolean> {
   });
 }
 
-export const DummyPlugin = {
-  run() {
-    return Promise.resolve([]);
-  },
-};
-
 // All of these functions exist to handle the fact that @glint/core won't be present as a peer dep
 // for non-Ember projects, so we need to lazily import and instantiate anything that depends on it
 // or else we get "module not found" errors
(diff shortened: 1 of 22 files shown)


chadhietala:
@egorio the thing we are currently doing is error prone and it just doesn't scale when new people come along to the codebase. If performance is a problem or to account for #584 or #585 we can re-visit but we're going to go ahead with this refactor.

lynchbomb:
I think this is a solid refactor `...register the constructors and let the queue executor be responsible for constructing them with the correct options` that is easy enough to create a mental model of this workflow for any new folks into the codebase. Strong approve. 

@egorio I agree WRT performance; however I think this refactor will be easier to implement any performance tweaks necessary.

egorio:
@chadhietala @lynchbomb the https://github.com/rehearsal-js/rehearsal-js/issues/584 is not about performance.

We currently have a possible issue because of the `files -> plugins` workflow.
We apply all plugins per file, then move to the next file.
At the same time, we expect fixes to be applied in a different that the current file.
Once we do this to the already processed file (different from the current one) we miss running other plugins (after -fix) again which can break formatting and reporting.

It's better to run `plugin on all files, then move to a different plugin`. 

> doesn't scale when new people come along to the codebase

I think using implementing this interface and creating an object directly `new` is easier to read and use, than learning the parent class implementation (super) and magic of how plugins are instantiated inside runner. We had kind of magic before, and I went from this one for the same reason - make it simpler + prepare for 584 :)


> register the constructors and let the queue executor be responsible for constructing them with the correct options

I did this while implementing the runner, but find that it doesn't make sense if we plan to run each plugin only once anyway.

packages/migrate/src/glint-utils.ts
```diff
@@ -106,20 +104,20 @@ export async function createGlintService(basePath: string): Promise<GlintService
   return new GlintService(glintCore, basePath);
 }
 
-export async function createGlintFixPlugin(): Promise<GlintFixPlugin> {
+export async function getGlintFixPlugin(): Promise<GlintFixPluginCtor> {
```
wondersloth, line 107: Just for clarification, changing this from `create` to `get` is because we're not instantiating something.

packages/service/src/plugin.ts
```diff
@@ -26,18 +43,69 @@ export interface PluginLogger {
...
+  queue<
+    PluginType extends Plugin = Plugin,
+    PluginOptionsType extends PluginOptions = PluginOptions
+  >(
+    plugin: PluginFactory<PluginType>,
+    ...optionsOrFilter:
+      | []
+      | [PluginOptionsType]
+      | [(fileName: string) => boolean]
+      | [PluginOptionsType, (fileName: string) => boolean]
```
wondersloth, line 89: This is hard to read. There is a lot of overloading and typing. Please add some documentation why this is necessary or simplify if at all possible.

packages/service/src/plugin.ts
```diff
@@ -82,15 +150,17 @@ export class PluginsRunner {
     fileName: string,
     allChangedFiles: Set<string>
   ): AsyncGenerator<Set<string>> {
-    for (const plugin of this.plugins) {
-      if (plugin.options.filter) {
-        if (!plugin.options.filter(fileName)) {
+    for (const level of this.levels) {
```
wondersloth, line 153: What is levels in this context of plugins. We could have `n` layers of plugins to apply?

Is this for multiple mixed contexts, glint first level, then JS next level?

PR #995, opened 2023-04-18
feat(cli): move command

lynchbomb:
This PR:

Adds a new command to rehearsal cli called `move`

```
Usage: rehearsal move|mv [options]

git mv conversion of JS files -> TS files

Options:
  -s, --source <path to source file | directory>  the explicit source file OR explicit directory with implicit child directories moved. migration strategy is ignored (default: "")
  -p, --childPackage <path to child package>      specify the child-package relative to process.cwd(). migration strategy is leveraged moving all necessary files in the dependency graph for this package (default: "")
  -d, --dryRun                                    Do nothing; only show what would happen (default: false)
  -h, --help                                      display help for command
```

Moving an explicit JS or GJS file with `--source <path-to-file>`:
file

Moving all JS or GJS files from a directory and sub-directories `--source <path-to-dir>`
dir

Moving all packages with migration-graph ordering `--childPackage <path-to-package>`
package

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -36,6 +36,9 @@
     "test:watch": "vitest --coverage --watch --config ./vitest.config.ts",
     "version": "pnpm version"
   },
+  "resolutions": {
+    "listr2": "^5.0"
+  },
   "dependencies": {
     "@rehearsal/migrate": "workspace:*",
     "@rehearsal/migration-graph": "workspace:*",
@@ -77,9 +80,6 @@
   "packageManager": "pnpm@8.2.0",
   "engines": {
     "node": ">=14.16.0"
-  },
-  "resolutions": {
-    "listr2": "^5.0"
   },
   "volta": {
     "extends": "../../package.json"
--- a/packages/cli/test/commands/migrate/analyze.test.ts
+++ b/packages/cli/test/commands/migrate/analyze.test.ts
@@ -15,7 +15,7 @@ import {
   prepareProject,
   sendKey,
 } from '../../test-helpers/index.js';
-import { TSConfig } from '../../../src/types.js';
+import type { TSConfig, MigrateCommandContext } from '../../../src/types.js';
 
 describe('Task: analyze', () => {
   let output = '';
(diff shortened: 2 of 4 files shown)


packages/cli/src/types.ts
```diff
@@ -2,12 +2,65 @@
...
+  source?: string;
+  dryRun: boolean;
+  basePath: string;
+};
+
+export type MoveCommandContext = {
+  skip: boolean;
+  strategy: MigrationStrategy | undefined;
+  input: string;
+  targetPackageAbs: string;
```
egorio, line 52: Don't see the usage of this one.

packages/cli/src/types.ts
```diff
@@ -2,12 +2,65 @@
...
+  workspaceRoot: string;
+  winstonLogger: Logger;
```
egorio, line 54: I would call it just `logger` with our own interface as we do in plugins:

```
export interface PluginLogger {
  log(message: string): void;
}
```

but can be updated next time.

packages/cli/src/types.ts
lynchbomb: yeah im going to do another full cleanup in future pr

packages/cli/src/types.ts
```diff
@@ -2,12 +2,65 @@
...
+  dryRun: boolean;
+  basePath: string;
+};
+
+export type MoveCommandContext = {
+  skip: boolean;
+  strategy: MigrationStrategy | undefined;
+  input: string;
+  targetPackageAbs: string;
+  workspaceRoot: string;
```
egorio, line 53: Don't see the usage of `targetPackageAbs`, `workspaceRoot`

packages/cli/src/types.ts
lynchbomb: im going to clean these types up in future pr

packages/cli/src/types.ts
lynchbomb: same going to clean this up in future pr

egorio (approved):

PR #1004, opened 2023-04-24
breaking(migrate): replaces migrate with fix

lynchbomb:
This is PR 3/4 of the CLI refactor of migrate and upgrade into fix. Essentially just to keep the PR's smallish.

!Smoke Test is expected to fail.

1. #1001 (Removes @rehearsal/upgrade and @rehearsal/cli upgrade)
2. #1002 (Refactor @rehearsal/reporter)
3. #1004 Replace migrate and upgrade with @rehearsal/fix and @rehearsal/cli fix
4. [Subsequent PR] Final Cleanup, Move Async/Worker, Remove Regen, Rename etc.

The new `fix` (aka migrate) workflow is:
```
  @rehearsal/fix workflow

  rehearsal fix <path to source file | directory>
    the explicit source file OR explicit directory with implicit child directories fixed.
    migration strategy is ignored

  rehearsal fix <path to child package> --graph --ignore tests/**/*
    specify the child-package relative to process.cwd(). migration strategy is leveraged fixing all necessary files in the dependency graph for this package.
    migration strategy is leveraged

  rehearsal fix <path to child package> --graph --devDeps
    same as above but devDependencies are considered in the graph
```

This PR is massive. So for review please filter the type ext to .ts. Critically for review please look at:

- Fix CLI Command: https://github.com/rehearsal-js/rehearsal-js/tree/9566a84815868c251ab294645b208f7fd9a15907/packages/cli/src/commands/fix
- Migrate: https://github.com/rehearsal-js/rehearsal-js/blob/9566a84815868c251ab294645b208f7fd9a15907/packages/migrate/src/migrate.ts
- Utils: https://github.com/rehearsal-js/rehearsal-js/blob/9566a84815868c251ab294645b208f7fd9a15907/packages/utils/src
- Flags/Args should align across the entire CLI
- Validate `migrate` and `cli fix` are consuming the `graph` API's correctly


egorio:
Please, consider next:

- Fix do not depend on existing js files (no more: " TS conversion task, no JS files detected")
- If no files are found - include the path in the error message, like Error "No files found in [path] instead of TS conversion task, no JS files detected
- Param --source to target particular file or directory for migration.

packages/cli/src/commands/fix/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    `specify the child-package relative to process.cwd(). migration strategy is leveraged moving all necessary files in the dependency graph for this package`,
+    ''
+  )
+  .option(
+    '-f, --format <format>',
+    'report format separated by comma, e.g. -f json,sarif,md,sonarqube',
+    parseCommaSeparatedList,
+    ['sarif']
+  )
+  .option('-w, --wizard', 'interactive wizard will prompt on a file-by-file basis', false)
```
chadhietala, line 62: I don't even know if we should do this right now.

packages/cli/src/commands/move/tasks/initialize.ts
```diff
@@ -18,7 +15,11 @@ export function initTask(
       ctx.childPackage = options.childPackage;
       // source file or dir should always exist within the basePath
       if (options.source) {
-        [ctx.jsSourcesAbs, ctx.jsSourcesRel] = validateSourcePath(options.basePath, options.source);
+        [ctx.jsSourcesAbs, ctx.jsSourcesRel] = validateSourcePath(
+          options.basePath,
+          options.source,
+          'js'
```
chadhietala, line 21: Can't you have `gjs` files as well?

packages/cli/src/commands/fix/index.ts
lynchbomb: yeah i keep going back/forth on this one. going file-by-file is nice to follow the type inference of rehearsal. i could still see this as valuable

packages/cli/src/commands/fix/tasks/initialize-task.ts
```diff
@@ -0,0 +1,86 @@
...
+      ctx.packageJSON = readJsonSync(basePath) as PackageJson;
+
+      // check if ember app/addon or glimmer project
+      ctx.projectType = 'base';
+      // if ember app or addon
+      if (isEmberApp(ctx.packageJSON) || isEmberAddon(ctx.packageJSON)) {
+        ctx.projectType = 'ember';
+      } else if (await isGlintProject(basePath)) {
+        ctx.projectType = 'glimmer';
+      }
```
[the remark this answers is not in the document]
lynchbomb: for the preflight check for glimmer apps. the tsconfig.json should have glint.env.'glimmerx` and dev deps are different

packages/cli/src/commands/move/tasks/initialize.ts
lynchbomb: yes. the `js` is just an identifier which will return all `*js` (right now its gjs, js)

packages/utils/src/cli.ts
```diff
@@ -738,3 +529,276 @@ export function findWorkspaceRoot(startDir: string = process.cwd()): string {
...
+    requiredTSConfig,
+    null,
+    2
+  )}`;
+
+  // !TODO NEED TO COMPARE AGAINST THE COMPILED TSCONFIG EG. EXTENDS WITH TS API
+  // grab all the top level keys from the requiredTSConfig
+  const missingKeys = Object.keys(requiredTSConfig).filter(
+    (key) => !Object.keys(tsConfig).includes(key)
+  );
```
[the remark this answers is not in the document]
lynchbomb: correct the `glint` key is root package only. however the `compilerOptions` can get inherited.

packages/cli/src/commands/fix/index.ts
chadhietala: We can keep the logic around but maybe leave this out for the time being. Just trying to limit the surface area of things that can go wrong.

packages/cli/src/commands/move/tasks/initialize.ts
lynchbomb: ill change the last arg for `validateSourcePath` to be either a glob or maybe array of extensions rather than obfuscate via
```
  const globPaths = fileType === 'js' ? ['.js,.gjs'] : ['.ts,.gts'];
```

packages/cli/src/commands/fix/index.ts
egorio: I agree to hide this. 

It seems useful, but I found it hard to use (when tried to use on voyager web) because it is hard to make a decision for a chunk of code without seeing whole files and other changes made before. I see this feature as IDE plugin :)

packages/cli/src/commands/fix/command.ts
```diff
@@ -0,0 +1,150 @@
...
+
+async function loadGraphTasks(): Promise<GraphTasks> {
+  return await import('../graph/tasks/graphOrderTask.js').then((m) => {
+    const { graphOrderTask } = m;
+
+    return {
+      graphOrderTask,
+    };
+  });
+}
```
chadhietala, line 150: We don't need to fix this now but we should create an issue that removes this type of stuff since we assume you installed the packages before hand.

packages/cli/src/commands/fix/tasks/initialize-task.ts
```diff
@@ -0,0 +1,103 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _ctx?: CommandContext
+): ListrTask<CommandContext, ListrDefaultRenderer> {
+  return {
+    title: `Initialize`,
+    task: async (ctx: CommandContext): Promise<void> => {
+      const { rootPath, graph } = options;
+
+      ctx.packageJSON = readJsonSync(resolve(rootPath, 'package.json')) as PackageJson;
+      ctx.projectType = 'base';
```
chadhietala, line 42: maybe `'js'` would be a better?

packages/cli/test/commands/fix/__snapshots__/fix.test.ts.snap
```diff
@@ -0,0 +1,312 @@
...
+Error: Failed to load parser @typescript-eslint/parser declared in .eslintrc.json: Cannot find module @typescript-eslint/scope-manager
+Require stack:
+- <tmp-path>/node_modules/@typescript-eslint/parser/dist/parser.js
+- <tmp-path>/node_modules/@typescript-eslint/parser/dist/index.js
+- /Users/imac/D/rehearsal-js/node_modules/.pnpm/@eslint+eslintrc@2.0.3/node_modules/@eslint/eslintrc/dist/eslintrc.cjs
+    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:995:15)
+    at Function.Module._load (node:internal/modules/cjs/loader:841:27)
+    at Module.require (node:internal/modules/cjs/loader:1067:19)
+    at require (node:internal/modules/cjs/helpers:103:18)
+    at Object.<anonymous> (<tmp-path>/node_modules/@typescript-eslint/parser/dist/parser.js:7:25)
```
chadhietala, line 304: Seems incorrect?

packages/service/src/glint-service.ts
```diff
@@ -91,6 +91,7 @@ export class GlintService implements Service {
     return this.service;
   }
 
+  // with-mapped-service.gts has 11 diagnostic glint
```
chadhietala, line 94: Delete?

packages/test-support/fixtures/ember/app-ts-template/app/components/ember-component.hbs
```diff
@@ -0,0 +1,28 @@
+<div ...attributes id={{unique-id}}>
+  {{this.required}}
+  {{this.optional}}
+  {{this.hasDefault}}
+
+  {{@required}}
+  {{@optional}}
+  {{@hasDefault}}
```
chadhietala, line 8: I'm kind of surprised that these don't have an annotation over them,

chadhietala (commented):
I left some comments. I think our best bet to validate this work is to locally try running it on some of our internal projects.

packages/cli/test/commands/fix/__snapshots__/fix.test.ts.snap
lynchbomb: yes this is known incorrect. i have this test skipped for now while i figure out whats up with the fixture. if you pull out this fixture from the mono-repo it runs fine without this error. looking into it

packages/cli/src/commands/fix/tasks/initialize-task.ts
lynchbomb: ill rename is 'base-ts'

packages/cli/src/prereqs.ts
wondersloth, line 1: This is great! 👍

packages/cli/test/commands/fix/__snapshots__/fix.test.ts.snap
```diff
@@ -0,0 +1,333 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Command: fix "base_ts_app" fixture > fix directory with src arg 1`] = `
+"info:    @rehearsal/fix<test-version>
+[STARTED] Initialize
+[SUCCESS] Initialize
+[STARTED] Infer Types
+[DATA] processing file: /vitest.config.ts
```
wondersloth, line 8: Is this expected?

packages/cli/test/fixtures/base_ts_app/base_ts_app.code-workspace
```diff
@@ -0,0 +1,24 @@
+{
```
wondersloth, line 1: Just curious.

What is this fixture used for?

wondersloth (approved):
Looks good.

Could've landed the ember TS app as a separate PR to shed some files off to review. That probably could've been rubber stamp committed.

packages/cli/test/fixtures/base_ts_app/base_ts_app.code-workspace
wondersloth: Drop. Just copied for a fixture. Not necessary.

packages/cli/test/commands/fix/__snapshots__/fix.test.ts.snap
lynchbomb: You saying `vitest.config.ts` <<<< this vitest file?

egorio (approved):
Land it. Will fix issues later, if they are there :)

PR #1005, opened 2023-04-24
chore: update README.md with new CLI Command API

wondersloth:
- Updated README with content from our internal TS Migration guide using new CLI commands.

# TODO
- [ ] Improve Typing. Update with correct regen / report command.
- [ ] Capture `--help` output from commands to add under Available Commands > Arguments
- [ ] Update Packages list.
- [ ] Add Setup guidance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ Rehearsal can be executed both manually and automated. Rehearsal creates a syste
 
 Ensure your project has the following dependencies:
 
+TBD
+
 # Installation
 
 Add `@rehearsal/cli` to your project as a **devDependency**.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,26 +9,35 @@
...
+
+Add `@rehearsal/cli` to your project as a **devDependency**.
 
 ```bash
 yarn add -D @rehearsal/cli
 # or
 pnpm add -D @rehearsal/cli
 ```
 
+Note: `@rehearsal/cli` must be installed as a devDependency as it requires your project's version of TypeScript to run.
```
egorio, line 38: Why this? It doesn't need to use the project's version to run, but it will be using a project version of TS to compile a Program (LanguageService).

wondersloth: Yes it does. How typescript is resolved is relative to the project's installed version of typescript. If `@rehearsal/cli` is installed globally or with volta install it will not work.

egorio: > If @rehearsal/cli is installed globally or with volta install it will not work.

Installed rehearsal is a bunch of JS files that use TS for some work.
If you install @rehearsal/cli globally (as I always do) it uses the closest to the target project TypeScript to do the work. 
If the project has TS installed as deps / devDeps - it will use it, if the project doesn't have TS installed - it will try to use global (which is not rehearsal issue, but project configuration).

egorio: Oh... seems it used to work, but broken now... 

```
Cannot find package 'typescript' imported from /Users/~/.volta/tools/image/packages/@rehearsal/cli/lib/node_modules/@rehearsal/cli/node_modules/@rehearsal/migrate/dist/src/migrate.js
```

weird

lynchbomb (approved):

PR #1011, opened 2023-04-27
chore: repro in app service bug

wondersloth:


egorio:
Fixed by https://github.com/rehearsal-js/rehearsal-js/pull/1013

PR #1013, opened 2023-04-28
feat: non-qualified service injection, based on provide map file

egorio:
Covers non-qualified service injection. 
The solution is using provided `services-map.json` file to include services by their names.

### DIFF
--- a/packages/migrate/test/migrate.test.ts
+++ b/packages/migrate/test/migrate.test.ts
@@ -222,6 +222,45 @@ export default class Hello extends Component {
       expect(report.summary[0].basePath).toMatch(project.baseDir);
     });
 
+    test('with service map', async () => {
+      const [inputs, outputs] = prepareInputFiles(project, ['with-mapped-service.gjs']);
+
+      const input: MigrateInput = {
+        basePath: project.baseDir,
+        sourceFiles: inputs,
+        entrypoint: '',
+        reporter,
+      };
+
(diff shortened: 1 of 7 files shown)


packages/migrate/test/migrate.test.ts
```diff
@@ -222,6 +222,45 @@ export default class Hello extends Component {
...
+        basePath: project.baseDir,
+        sourceFiles: inputs,
+        entrypoint: '',
+        reporter,
+      };
+
+      for await (const _ of migrate(input)) {
+        // no ops
+      }
+      const expected = `// @ts-expect-error @rehearsal TODO TS2307: Cannot find module 'services/moo/moo' or its corresponding type declarations.
```
chadhietala, line 238: This is because we don't have the [declaration here](https://github.com/rehearsal-js/rehearsal-js/pull/1013/files#diff-4b9ecf2ee31ec65529f94a12d5291120d70750645dd279730335bd2a5a334b0eR11) correct?

chadhietala (approved):
Just a clarifying question otherwise this looks good.

packages/migrate/test/migrate.test.ts
egorio: Yeah. Want to cover a case when service is in the map, but TS can't load it.

lynchbomb (approved):

PR #1021, opened 2023-05-02
fix: get diagnostics workflow

chadhietala:
This fixes the workflow that has been failing since https://github.com/rehearsal-js/rehearsal-js/pull/965 was introduced.


lynchbomb (approved):

lynchbomb:
Can we add a status indicator for this job? I didn’t notice it was failing.

chadhietala:
@lynchbomb I believe the status indicators only work on pull_request and this thing only runs on pushes to master.

PR #1025, opened 2023-05-03
`graph` and `mv` graph intergration

chadhietala:
This PR reworks both `graph` and `move` (they rely on one another) to allow for subsetting a project's package and module graph. Consider the following:

## `rehearsal graph`

`graph` is intended to be used to figure out reverse topsort order of both packages (entities which contain modules) and modules aka files. In the simplest form `graph` can just give you a list of  reverse topsorted modules at a given path. However, a developer may want to know if they are trying to migrate a package that has dependencies that have yet to be migrated. To account for this use case `graph` needed to be able to follow its dependencies to give the developer a fully materialized sub-graph in leaf to root order.  By migrating from leaf-most to root-most, the types that get inferred become richer as you migrate towards the root.

To account for this we have changed the CLI to have the following interface:

`rehearsal graph ./somePath`: Prints just the files in the package in top sorted order
`rehearsal graph ./somePath --deps`: Subsets the graph and only follows packages in "dependencies"
`rehearsal graph ./somePath --devDeps`: Subsets the graph and only follows packages in "devDependencies"
`rehearsal graph ./somePath --ignore some-package`: Only valid when --deps and/or --devDeps are present. Allows to ignore packages from analysis

## `rehearsal move`

`move` in its simplest form takes a path and `git mv`s the javascript files at that path to typescript files by just changing the extension. However, `move` can also use the same functionality as `graph` to move whole graphs of files.

To account for this we have changed the CLI to have the following interface:

`rehearsal move ./somePath`: Moves just the files in the path
`rehearsal move ./somePath --graph --deps` Follows the deps of ./somePath and moves the subgraph
`rehearsal move ./somePath --graph --devDeps`: Follows the devDeps of ./somePath and moves the subgraph
`rehearsal move ./somePath --graph --deps --devDeps`: Follows both and moves both
`rehearsal move ./somePath --graph --deps --ignore some-other-package`: Ignore can be added to further subset

## Testing

The smoke test is going to fail because the tests in [that repo](https://github.com/rehearsal-js/smoke-test) need to be updated there to account for these changes. Ideally though we just collapse that repo into this repo under its own workspace.

## Follow Ups

- https://github.com/rehearsal-js/smoke-test/pull/16
- https://github.com/rehearsal-js/rehearsal-js/issues/1026
- https://github.com/rehearsal-js/rehearsal-js/issues/1027


packages/cli/src/commands/graph/tasks/graphOrderTask.ts
```diff
@@ -23,22 +23,35 @@ export function graphOrderTask(
...
 
       if (process.env['TEST'] === 'true') {
         // Do this on the main thread because there are issues with resolving worker scripts for worker_threads in vitest
         const { intoGraphOutput } = await import('./graphWorker.js').then((m) => m);
         const { getMigrationOrder } = await import('@rehearsal/migration-graph').then((m) => m);
         // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-call
-        order = intoGraphOutput(getMigrationOrder(basePath), basePath);
+        order = intoGraphOutput(
+          getMigrationOrder(srcDir, {
+            basePath,
```
egorio, line 36: How about, instead of `basePath` (root package for monorepos) use `rootPath` (always root or repo) and `packagePath` (or `[workspacePath](https://docs.npmjs.com/cli/v7/using-npm/workspaces)`) for the current target package?

For single-repo `packagePath` will be equals to `rootPath`... but it will be clear where do we expect to use root (report, configs) and where current package dir.

@chadhietala @lynchbomb ?

chadhietala: Yea I'm not happy with the naming of this stuff. I'm going to come back around and cleanup this stuff once the tests are green.

lynchbomb (approved):

PR #1030, opened 2023-05-05
feat: graph --ignore with globs

chadhietala:
This PR adds glob/file support to the `--ignore` flag on `graph` and `move`. By default `graph` will show all reachable js and ts files in a project, where as `move` will only work on graphs of ts files.


chadhietala:
@lynchbomb lets look into the failures on Monday. Currently the master branch is failing and it has to do with the fact that we are npm installing deps and there seems to be some issues with typescript + @tyscript-eslint.

chadhietala:
@lynchbomb NM I figured it out and fixed it in #1033. This is now G2G.

lynchbomb (approved):

PR #1044, opened 2023-05-12
test(smoke-test): moves org smoke-test into a package

lynchbomb:
This PR is largely just a file move from https://github.com/rehearsal-js/smoke-test into https://github.com/rehearsal-js/rehearsal-js/tree/master/packages/smoke-test with:

- updates the smoke-test to included the updated CLI with (Fix | Move | Graph)
- all of the fixtures are pre-configured, which means we can rely on `pnpm` and dont need `npm`
- all of the devDeps are linked from the smoke-test package. a downside to this is `pnpm install` at the root take a little longer
- the only @rehearsal dep in the smoke-test is `@rehearsal/cli`
- the entire smoke-test runs in a separate GH-Action execution under `test:slow`; this allows for faster test execution time and unblocks local development when running `test` 

### Future State (didn't do)

- the fixtures in this package should be the source of truth going forward and we can remove the duplication from other packages
- likely a bunch of end-to-end tests can be removed for smoke-test coverage (thinking packages/cli and packages/migrate) 
- https://github.com/rehearsal-js/smoke-test can be archived


egorio (approved):

wondersloth (approved):
Looks good.

Some things I noticed it looks like super rentals is bootstrapped already with the required dependencies?

This way we don't have to worry about ember generate ember-cli-typescript and tsconfig setup, so we can use pnpm all the way through. 

Also we're dropping 3.28 and 4.4 tests, for latest ember 4.11 👍

lynchbomb:
> Some things I noticed it looks like super rentals is bootstrapped already with the required dependencies
@wondersloth yep!

PR #1046, opened 2023-05-15
fix: migrate files from selected source path

egorio:
We have 2 types of paths: `root` and `src`.
Need to use `src` for find the correct tsconfig.json and files that need to be migrated.

### DIFF
--- a/packages/cli/src/commands/fix/command.ts
+++ b/packages/cli/src/commands/fix/command.ts
@@ -40,7 +40,7 @@ fixCommand
   .alias('infer')
   .name('fix')
   .description('fixes typescript compiler errors by inferring types on .*ts files')
-  .argument('[srcDir]', 'path to directory containing a package.json', process.cwd())
+  .argument('[targetPath]', 'path to file or directory to migrate', process.cwd())
   .option('-g, --graph', 'enable graph resolution of files to move', false)
   .option('--devDeps', `follow packages in 'devDependencies'`, false)
   .option('--deps', `follow packages in 'dependencies'`, false)
@@ -57,16 +57,15 @@ fixCommand
     ['sarif']
   )
   .addOption(
-    new Option('--rootPath <project base path>', '-- HIDDEN LOCAL DEV TESTING ONLY --')
+    new Option('--rootDir <project root directory>', '-- HIDDEN LOCAL DEV TESTING ONLY --')
       .default(process.cwd())
-      .argParser(() => process.cwd())
       .hideHelp()
   )
-  .action(async (src: string, options: FixCommandOptions) => {
-    await fix(src, options);
+  .action(async (targetPath: string, options: FixCommandOptions) => {
+    await fix(targetPath, options);
   });
 
-async function fix(src: string, options: FixCommandOptions): Promise<void> {
+async function fix(targetPath: string, options: FixCommandOptions): Promise<void> {
   winstonLogger.info(`@rehearsal/fix ${version?.trim()}`);
   if (options.graph && !options.deps && !options.devDeps) {
     console.warn(
@@ -86,7 +85,7 @@ async function fix(src: string, options: FixCommandOptions): Promise<void> {
     throw new Error(`'--ignore' can only be passed when you pass --graph`);
   }
 
-  if (!src) {
+  if (!targetPath) {
     throw new Error(`@rehearsal/fix: you must specify a package or path to move`);
   }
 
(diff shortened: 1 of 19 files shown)


lynchbomb (commented):
we already have source path as the “fix” command argument and rootPath can be set as well with the flag.

ex. “rehearsal fix ARG-SRC-PATH —-rootPath ./“

packages/cli/src/commands/fix/command.ts
```diff
@@ -105,9 +105,9 @@ async function fix(src: string, options: FixCommandOptions): Promise<void> {
           ignore: options.ignore,
           skipPrompt: true,
         }),
-        convertTask(options),
+        convertTask(src, options),
```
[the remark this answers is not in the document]
lynchbomb: We should actually freeze the ctx object in init task

lynchbomb:
@egorio something does not seem correct with this implementation. what is the failing test case this is fixing / repro details?

packages/migrate/src/migrate.ts
```diff
@@ -79,28 +82,28 @@ export async function* migrate(input: MigrateInput): AsyncGenerator<string> {
   const { fileNames: someFiles } = parseJsonConfigFileContent(
     configFile,
     ts.sys,
-    basePath,
+    srcPath,
```
lynchbomb, line 85: `srcPath` here can be a directory path OR a file path. however `parseJsonConfigFileContent` expects only a directory path.

lynchbomb (changes requested):
left a few comments

packages/cli/src/commands/fix/command.ts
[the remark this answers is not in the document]
egorio: I did srcDir as a part of ctx, but then realized that srcDir is differentiated from options in all other tasks function ([one](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/graph/tasks/graphOrderTask.ts#LL17C3-L17C18), [two,](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/move/tasks/initialize.ts#L10) [three](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/move/tasks/mv.ts#L14)), and even in CLI it's coming as an [argument](argument).  

Just decided to follow the same principal.

packages/migrate/src/migrate.ts
egorio: That's interesting. 
It sound like we expect srcDir to be a directory... while I also expected to pass path to file as well :

`[srcDir]', 'path to directory containing a package.json`

egorio:
Use `targetPath` (will probably rename with targetSrcPath` in the future) instead or `src`... because it's still unclear inside the code if src is a whole project or just some part.

And `packageDir` instead of `childRootDirAbs` for the same reason. Child sounds confusing when there is a simple project without subpackages)

packages/migrate/test/__snapshots__/migrate.test.ts.snap
```diff
@@ -14,7 +14,7 @@ export default class Hello extends Component {
 `;
 
 exports[`fix > .gts > when missing a local prop 1`] = `
-"import Component from \\"@glimmer/component\\";
```
egorio, line 17: This will be addressed next PR

packages/migrate/src/migrate.ts
```diff
@@ -136,7 +136,7 @@ export async function* migrate(input: MigrateInput): AsyncGenerator<string> {
     .queue(
       LintPlugin,
       {
-        eslintOptions: { cwd: basePath, useEslintrc: true, fix: true },
+        eslintOptions: { cwd: packageDir, useEslintrc: true, fix: true },
```
wondersloth, line 139: `packageDir` here is the targetPath derived packageDir?

When these plugins run, why is it important to pass the dir to these plugins?

packages/migrate/test/__snapshots__/migrate.test.ts.snap
wondersloth: Oh this is a prettier config issue where it's not recognizing the correct config because we are passing `packageDir` to the plugins?

For each case of a plugin, should be findingup to find .eslintrc or .prettierrc?

packages/utils/src/cli.ts
```diff
@@ -405,6 +405,15 @@ export function readTSConfig<T>(configPath: string): T {
...
 
+/**
+ * Finds the closest ancestor to `startPath` directory containing package.json file
+ */
+export function findPackageRootDirectory(startPath: string): string | undefined {
+  const packageJSONPath = findupSync('package.json', { cwd: startPath });
+
+  return packageJSONPath ? parse(packageJSONPath).dir : undefined;
+}
+
```
wondersloth, line 416: Refactor ember-app-project-graph to use this method.

Also had a stopping point.

```
    // Findup till we find package.json relative to the dir of the entrypoint.
    const foundPackageJson = findUpSync('package.json', {
      cwd: lookUpDir,
      stopAt: this.rootDir,
    });
```

packages/migrate/src/migrate.ts
egorio: Correct. Eslint is looking for its config starting from cwd value

packages/utils/src/cli.ts
egorio: Will do that, thanks.

egorio:
> Oh this is a prettier config issue where it's not recognizing the correct config

Prettier is finding its config correctly. 
Prettier formatting is failing due to the parser throwing an error.

Planning to solve this next PR

PR #1051, opened 2023-05-17
fix: adjust path used by glint service

egorio:
Misused root and package path in my previous PR.


packages/cli/src/commands/fix/tasks/convert-task.ts
```diff
@@ -38,7 +38,7 @@ export function convertTask(
         commandName: '@rehearsal/fix',
       });
 
-      const packageDir = findPackageRootDirectory(targetPath) || rootPath;
+      const packageDir = findPackageRootDirectory(targetPath, rootPath) || rootPath;
```
egorio, line 41: Based on @wondersloth [suggesition](https://github.com/rehearsal-js/rehearsal-js/pull/1046#discussion_r1195772314)

packages/utils/src/cli.ts
```diff
@@ -2,7 +2,7 @@ import { dirname, join, normalize, isAbsolute, relative, resolve, extname, parse
 import { existsSync, readFileSync } from 'node:fs';
 import { readJSONSync, writeJSONSync } from 'fs-extra/esm';
 import { compare } from 'compare-versions';
-import findupSync from 'findup-sync';
+import { findUpSync } from 'find-up';
```
wondersloth, line 5: nit: findupSync

findup is treated as one lower case word.
https://www.npmjs.com/package/findup-sync

wondersloth (commented):
Rename back to findupSync

This will slim the changes, and align to how the is reference in the docs.

Otherwise approve.

wondersloth (changes requested):
Can you refactor this method:

https://github.com/rehearsal-js/rehearsal-js/blob/e1128c15b7965cb4f0251ff02f58dcea6d18e0a6/packages/migration-graph-ember/src/entities/ember-app-project-graph.ts#L154

To use your new util method.

packages/utils/src/cli.ts
egorio: `findupSync` is from `findup-sync` and I don't see stopAt there.
  while `findUpSync` is form `find-up`

lynchbomb (approved):
Be sure to validate the fix before merging

lynchbomb:
``"import type FooService from \\"foo/services/foo-service\\"`

This type **should** be inlined with the codefix; the snapshot is throwing about it

lynchbomb (changes requested):
Once the snapshot / test is resolved this is good

egorio:
> This type should be inlined with the codefix; the snapshot is throwing about it

Actually, the failure is legit because the addon knows nothing about those services (because they are not included).
I added an additional case where an addon / subpackage is importing required types and has no comments added.

egorio:
> To use your new util method.

Done

packages/migrate/src/migrate.ts
```diff
@@ -1,5 +1,5 @@
 import { resolve, join } from 'node:path';
-import findup from 'findup-sync';
+import { findUpSync } from 'find-up';
```
wondersloth, line 2: You can dry this up.

packages/service/src/rehearsal-service-host.ts
```diff
@@ -5,7 +5,7 @@ import ts, {
   LanguageServiceHost,
 } from 'typescript';
 import { addDep } from '@rehearsal/utils';
-import findupSync from 'findup-sync';
+import { findUpSync } from 'find-up';
```
wondersloth, line 8: You can consolidate all findUp usage with some util methods.

```
findClosestPackageJson(somePath, stopat) {
  findClosestFile('package.json',  dirname(somePath), stopAt);
}

findClosestFile(filename, start, stop) {
 // We can use this for all findup logic, and have helper methods on top for 
 // The prettier helper can use this. This way we have one findup dependency on util and we don't have to coordinate across all these packages.
}
```

wondersloth (changes requested):

wondersloth (approved):

lynchbomb (approved):

PR #1053, opened 2023-05-17
Module Graph

chadhietala:
This PR is a re-think of how we crawl a project to determine the order in which we need to fix or move files. It moves us from a package level graph to solely a module level graph e.g. following and resolving `import` statements.

**Points of Interest:**

Since this PR has a lot of things going on it due to moving functionality around, here are the important bits:

- [Resolver](https://github.com/rehearsal-js/rehearsal-js/pull/1053/files#diff-5d197c72b22b4bb5388f345b4b64232cf84c1f3903480f44549d72344d4cd250): Responsible for resolving imports
- [PackageGraph](https://github.com/rehearsal-js/rehearsal-js/pull/1053/files#diff-82a85f67f77f44c1ac049d20abf47fbc095c97a788d4e19d752ee15891666779): The actual graph implementation
- [GraphViz Output](https://github.com/rehearsal-js/rehearsal-js/pull/1053/files#diff-5e0e582729b93050bf1beaae832f219beeda92adcbfbf1b8fdd66471446f43a2): Generates a graph like the following

Screenshot 2023-05-18 at 3 02 22 PM

## Background

Below outlines the shortcomings of our current approach to graphing files in a project.

### Package Graph Only

Prior to the changes we attempted to create entities within the graph that were framework specific graph projects, graphs, and packages. These entities, specifically the Ember related ones, would hardcode resolution rules like:

1) look for an `index.js` with `moduleName` method or `name` to see what the package name is used in code
2) look for `ember-addon` keyword in package.json and include/exclude certain files
3) look for  `ember-addon` field in package.json for `paths` for in-repo addons

These conventions and configurations allowed use to find packages but *did not* follow the actual es module graph. This means that when we would cross a module boundary we would only glob the files in that package and then traverse this graph. The end result was a list of topographically sorted packages with a list of topographically sorted files in those packages. For example:

Given code that looks like:

```
// lib/some-addon/addon/utils/c.ts
import Foo from 'foo/baz';
```

```ts
// lib/some-other-addon/addon/utils/baz.ts
export const baz = 'baz';
```

And a file system that looked like this:

```
lib
├── some-addon
│   └── addon
│       ├── a.ts
│       └── utils
│           ├── c.ts
│           └── d.ts
└── some-other-addon
    └── addon
        ├── b.ts
        └── utils
            ├── baz.ts
            └── bizz.ts
```

What you would get back as a result would be something like:

```ts
{
  packages: [
     {
        name: 'some-other-addon',
        files: [
          'lib/some-other-addon/addon/utils/baz.ts',
          'lib/some-other-addon/addon/utils/bizz.ts'
          'lib/some-other-addon/addon/b.ts'
        ]
     },
     {
        name: 'some-addon',
        files: [
          'lib/some-addon/addon/a.ts',
          'lib/some-addon/addon/utils/d.ts'
          'lib/some-addon/addon/utils/c.ts'
        ]
     }
  ]
}
```

As you can see we would correctly top sort the package graph and even the files within that package, however we were not follow just the import graph leaving us with way more files than were actually need to migrate. While this is not fatal, it does mean that showing the true graph of files is lost.

### Reliance on `devDependencies` and `dependencies` and `--deps` and `--devDeps` flags

One of the ways we attempt to find packages is to first glob all `package.json`s in a project at the [beginning of graph traversal](https://github.com/rehearsal-js/rehearsal-js/blob/11e26ed11dd2b970057ae607210cc2499cc35a21/packages/migration-graph-shared/src/entities/project-graph.ts#L254) to build a mapping of all packages within a project. From there we start resolving files using this pre-computed map to draw [edges between packages](https://github.com/rehearsal-js/rehearsal-js/blob/11e26ed11dd2b970057ae607210cc2499cc35a21/packages/migration-graph-ember/src/entities/ember-app-project-graph.ts#L361) whenever code from outside of the package is referenced. When you point `graph`, `move` or `fix` at some directory or file and you typically had add `--deps` and/or `--devDeps` flags to limit or expand the resolution of packages based on the type of code you were move e.g. tests vs app code. 

While this worked to narrow the resolution of files it was simply a work around for not being able to follow the module graph it self.

### Graph was opt-in

Since graphing a project pulled in more files than were necessary you had to opt-in `move` and `fix` into graph following with the `--graph` flag.

### Glob Heuristics

Since we were only following the package graph and not the module graph, we had to maintain a set of [framework specific glob patterns](https://github.com/rehearsal-js/rehearsal-js/blob/23a816bddc3961c7f4d45f6f1c67950fd8ce7bae/packages/migration-graph-ember/src/utils/excludes.ts#L14). While this worked, it was a symptom of the larger problem of not following the module graph.

### Performance
Due to the non-precise nature of following files across package boundary we would process way more files than needed which lead to slower resolution of the graph. In addition that we were using [dependency-cruiser](https://github.com/sverweij/dependency-cruiser) which does a [full parse](https://github.com/sverweij/dependency-cruiser/blob/aaf4e53e86ef11e09b4b6e957f6fe81243623901/src/extract/parse/to-javascript-ast.mjs#L29) of all files even though all we care about are import declarations.

## About This PR

This PR does the following:

- You can now pass any file path or directory path to `graph`, `move` or `fix`. If you pass a file path it will use that as the root most node of the graph and walk all the way to the leafs. If a directory is passed it will glob the files at the directory path and then walk to their leafs. If files in the directory make references to one another edges will be put between them
- Removes the need to pass `--graph`, `--deps`, `--devDeps`. Graph traversal is on by default. To opt-out of graphing out the project we have add `--no-graph`
- Instead of doing the heuristics based resolution where we look for Ember specific configuration or globing all `package.json`s up front, we **rely** on [tsconfig's paths](https://www.typescriptlang.org/tsconfig#paths) to resolve files that exist in other parts of the code base. [Example](https://github.com/rehearsal-js/rehearsal-js/blob/c6c752703be83c5d9d0d758626ff386f88c24f3d/packages/migration-graph/test/resolver.test.ts#L125)
- Instead of using dependency-cruiser we are using a wasm based spec compliant [module lexer](https://github.com/guybedford/es-module-lexer) to find imports. In cases of `jsx`, `tsx`, `gts`, and `gjs` we first pre-process with [swc or ember-template-imports](https://github.com/rehearsal-js/rehearsal-js/blob/c6c752703be83c5d9d0d758626ff386f88c24f3d/packages/migration-graph/src/resolver.ts#L120) to arrive at code that can go through the lexer. Subgraphs are now sub 500ms and massive projects (5m loc +) take less than 2 min to resolve.
- You can now produce graphviz output by passing `graph` an `--output` that points to a file with a `.dot` extension


packages/migration-graph/src/discover-services.ts
```diff
@@ -0,0 +1,295 @@
...
+  ImportSpecifier,
+  ModuleItem,
+  NamedImportSpecifier,
+  StringLiteral,
+} from '@swc/core';
+import type { CustomImportResolver } from './types.js';
+
+const DEBUG_CALLBACK = debug('rehearsal:migration-graph:discover-ember-service-dependencies');
+
+export function discoverServiceDependencies(
```
egorio, line 23: 👍 💯

packages/cli/src/commands/fix/command.ts
```diff
@@ -61,46 +62,23 @@ fixCommand
       .default(process.cwd())
       .hideHelp()
   )
-  .action(async (targetPath: string, options: FixCommandOptions) => {
-    await fix(targetPath, options);
+  .action(async (srcPath: string, options: FixCommandOptions) => {
```
egorio, line 65: could you please use targetSrcPath... just trying to make it clearer down the code.

packages/cli/src/commands/fix/tasks/convert-task.ts
```diff
@@ -19,7 +18,7 @@ export function convertTask(
     title: 'Infer Types',
     task: async (ctx: CommandContext, task): Promise<void> => {
       const { rootPath, ignore } = options;
-      const { projectName, sourceFilesAbs } = ctx;
+      const { projectName, orderedFiles } = ctx;
```
egorio, line 21: I also find that using context is like using global var, very hard to keep mental model of the app in the memory while debugging. 

I would prefer to pass the result of one task to the next one input... but probably not in this PR :)

packages/cli/src/utils/prereq-checks.ts
```diff
@@ -0,0 +1,244 @@
...
+  if (!existsSync(packageJsonPath)) {
+    throw new Error(
+      `${packageJsonPath} does not exists. Please run rehearsal inside a project with a valid package.json.`
+    );
+  }
+  return true;
+}
+
+export function isExistsTSConfig(basePath: string): boolean {
+  const tsConfigPath = resolve(basePath, 'tsconfig.json');
```
egorio, line 140: I think this: https://github.com/rehearsal-js/rehearsal-js/blob/6363ff411e8b24cd1badd62bd4de3e437218a111/packages/migrate/src/migrate.ts#L69-L73

egorio: https://github.com/rehearsal-js/rehearsal-js/issues/1056

packages/cli/src/commands/graph/tasks/graphOrderTask.ts
```diff
@@ -1,11 +1,14 @@
...
-import debug from 'debug';
+import { existsSync } from 'node:fs';
 import { ListrDefaultRenderer, ListrTask } from 'listr2';
-import type { PackageEntry, CommandContext, GraphTaskOptions } from '../../../types.js';
-
+import { discoverServiceDependencies, Resolver, topSortFiles } from '@rehearsal/migration-graph';
+import { readJSON } from '@rehearsal/utils';
+import debug from 'debug';
+import { getFiles, writeOutput } from './graphWorker.js';
+import { Response } from './types.js';
```
lynchbomb, line 10: import type { Response } from './types.js';

packages/cli/src/commands/graph/tasks/graphOrderTask.ts
```diff
@@ -71,51 +89,15 @@ export function graphOrderTask(
...
       }
 
-      // set the order on the context so we can use it in other tasks
-      ctx.migrationOrder = order;
-
-      // skip the output and prompt. for when we just want the order
-      if (options.skipPrompt) {
-        ctx.sourceFilesRel = order.packages.flatMap((pkg) => pkg.files);
-        ctx.sourceFilesAbs = ctx.sourceFilesRel.map((file) => resolve(rootPath, file));
+      if (!output && !skipPrompt) {
```
lynchbomb, line 92: `!skipPrompt` now that the prompt is removed we can remove this conditional and wherever its being set

packages/cli/src/commands/graph/tasks/graphWorker.ts
```diff
@@ -1,66 +1,100 @@
...
+  Resolver,
+  topSortFiles,
+  discoverServiceDependencies,
+  getExcludePatterns,
+  SUPPORTED_EXTENSION,
+  PackageGraph,
+  generateDotLanguage,
+} from '@rehearsal/migration-graph';
+import FastGlob from 'fast-glob';
+import { GraphWorkerOptions } from './types.js';
```
lynchbomb, line 14: import type { GraphWorkerOptions } from './types.js';

packages/cli/test/commands/fix/__snapshots__/fix.test.ts.snap
```diff
@@ -272,144 +276,80 @@ export function generateRandomGrid(
...
+[DATA] processing file: /app/components/bar/index.ts
+[DATA] processing file: /app/components/test-cases/subclassing/parent.ts
+[DATA] processing file: /app/components/test-cases/subclassing/child.ts
+[DATA] processing file: /app/pods/components/baz/component.ts
 [TITLE] Types Inferred
-[TITLE]   38 errors caught by rehearsal
-[TITLE]   9 have been fixed by rehearsal
-[TITLE]   29 errors need to be fixed manually
-[TITLE]     -- 27 ts errors, marked by @ts-expect-error @rehearsal TODO
+[TITLE]   22 errors caught by rehearsal
```
lynchbomb, line 305: Will you double check the reduction in errors caught and fixes is accurate?

packages/cli/test/commands/fix/init-task.test.ts
```diff
@@ -278,20 +278,16 @@ describe('Fix: Init-Task', () => {
...
       format: ['sarif'],
-      graph: false,
-      devDeps: false,
-      deps: false,
       ignore: [],
     };
     const tasks = [initTask(src, options)];
     const ctx = await listrTaskRunner<CommandContext>(tasks);
-    const sanitizedAbsPaths = ctx.sourceFilesAbs?.map((path) => {
+    const sanitizedAbsPaths = ctx.orderedFiles?.map((path) => {
```
lynchbomb, line 285: should we append `rel/abs` on this >> `ctx.orderedFiles`

lynchbomb (changes requested):

lynchbomb (approved):

NullVoxPopuli:
idk if it helps for gjs/gts, but we have these tools now:
- github.com/NullVoxPopuli/ember-estree - also does hbs? I forget, tbh
- github.com/NullVoxPopuli/zmod-ember

Tho, I see you're using: https://github.com/guybedford/es-module-lexer now
I'm using that here: https://github.com/NullVoxPopuli/limber/pull/1995

so maybe you don't need anything else

PR #1057, opened 2023-05-18
chore: delete regen

chadhietala:
Prior to reorienting the project around `graph`, `move`, `fix`, we had `regen` that would regenerate the report that contained all places where there are TODOs. This independent concept is no longer needed because `fix` can be reran and produce a report with the same information.

Removing this whole package 🫡.


lynchbomb (approved):
✌️

PR #1058, opened 2023-05-18
chore: fix base-codefixes test

egorio:
Currently, test for base codefixes do nothing and snapshots contain original (not modified) files contents... this is why they pass. 

This PR is fixing those tests.


packages/migrate/src/glint-utils.ts
```diff
@@ -1,41 +1,10 @@
...
-// The list of extensions that we expect to be handled by Glint{Fix,Check} plugins. Note that
-// in any ember/glimmer project, we'll use the glint *service* for all files. This list is only
-// indicating which extensions are handled by the plugins
-export const GLINT_EXTENSIONS = ['.gts', '.hbs'];
-
-// The list of dependencies we look for to determine if we're in a glint project. If we find one
-// of these, we use glint. Otherwise, we use the regular Rehearsal service
-export const GLINT_PROJECT_FILES = ['ember-source', '@glimmer/component', '@glimmerx/component'];
-
-export async function shouldUseGlint(basePath: string): Promise<boolean> {
```
egorio, line 21: Clean up

wondersloth: Why can we remove this?

egorio: It's not used.

egorio: We have same in @service

wondersloth (approved):

Issue #1060, opened 2023-05-18
bug: annotateWithStrictTypeFromJSDoc breaks and fails with `Cannot read properties of undefined (reading 'flags')`


wondersloth:
Found on an internal MP when testing against `learning-web`.

## Affected Version
**3.0.2-beta**

## Steps to reproduce

Checkout `learning-web` 

```
$ yarn rehearsal fix lib/core/addon/components/entity-carousel.ts
info:    @rehearsal/fix 3.0.2-beta
✔ Initialize
⠴ Infer Types
  › processing file: /addon/components/entity-carousel.ts
✖ Cannot read properties of undefined (reading 'flags')
```
The program fails at:
https://github.com/rehearsal-js/rehearsal-js/blob/b27674fff2d0f12630ef6826750df11999d19ffc/packages/ts-utils/src/helpers/typescript-ast.ts#L130

`typeNode.typeName.symbol` is `undefined`

```
  /**
   * Given a start and end index (and optional array of entities), return an `EmberArray`
   * of object containing the entity's `content`, `isLoading` state, and `cumulativeIndex`.
   * @private
   * @type {Function}
   * @param {[number, number]} indices - start and end index of entities to get
   * @param {Array<Entity>} entities - array of entities to transform
   * @return {Array<{content: Entity | null, isLoding: boolean, cumulativeIndex: number}>} an array of objects
   */
  _getEntitiesWithLoadingState([startIndex, endIndex], entities = this.entities) {
    const newEntities = emberArray();
    for (let index = startIndex; index <= endIndex; index++) {
      const content = entities.objectAt(index);
      newEntities.pushObject(
        EmberObject.create({
          isLoading: !content,
          content,
          cumulativeIndex: index,
        })
      );
    }
    return newEntities;
  }
```

If we remove / ignore the comment the file can be parsed.

## Stack Trace
```
getMembersOfSymbol (/node_modules/typescript/lib/typescript.js:54848)
getTypeFromTypeLiteralOrFunctionOrConstructorTypeNode (/node_modules/typescript/lib/typescript.js:59630)
getTypeFromTypeNodeWorker (/node_modules/typescript/lib/typescript.js:59998)
getTypeFromTypeNode (/node_modules/typescript/lib/typescript.js:59924)
map (/node_modules/typescript/lib/typescript.js:259)
typeArgumentsFromTypeReferenceNode (/node_modules/typescript/lib/typescript.js:57458)
[...]

PR #1061, opened 2023-05-18
fix: catch and return false if getTypeFromTypeNode fails

wondersloth:
Fixes: #1060 

Validated locally.

- [x] Needs a test.

### DIFF
--- a/packages/codefixes/test/__snapshots__/base-codefixes.test.ts.snap
+++ b/packages/codefixes/test/__snapshots__/base-codefixes.test.ts.snap
@@ -563,6 +563,20 @@ export function think(
 export function thinkTwice(a) {
   return a;
 }
+
+/**
+ * The return type is throwing a error
+ * @param {[number, number]} ab
+ * @param {Array<string>} c
+ * @return {Array<{content: string}>}
+ */
+// @ts-expect-error @rehearsal TODO TS7031: Binding element 'a' implicitly has an 'any' type.
+export function moreCases([a, b], c: Array<string>) {
+  a;
+  b;
+  c;
+  return c;
+}
 "
 `;
 
--- a/packages/codefixes/test/__snapshots__/base-codefixes.test.ts.snap
+++ b/packages/codefixes/test/__snapshots__/base-codefixes.test.ts.snap
@@ -565,17 +565,36 @@ export function thinkTwice(a) {
 }
 
 /**
- * The return type is throwing a error
- * @param {[number, number]} ab
+ * The return type is throwing an error
+ * @param {[number, number]} a-and-b
  * @param {Array<string>} c
  * @return {Array<{content: string}>}
  */
 // @ts-expect-error @rehearsal TODO TS7031: Binding element 'a' implicitly has an 'any' type.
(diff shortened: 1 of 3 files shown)


packages/ts-utils/src/helpers/typescript-ast.ts
```diff
@@ -129,7 +129,14 @@ export function isVariableOfCatchClause(node: Node): boolean {
 
 export function canTypeBeResolved(checker: TypeChecker, typeNode: TypeNode): boolean {
   if (isTypeReferenceNode(typeNode)) {
```
egorio, line 131: You could change this if statement to 

```
if (isTypeReferenceNode(typeNode)) {
    const typeArguments = typeNode.typeArguments || [];

    const isTypeNodeStrict = (checker: TypeChecker, typeNode: TypeNode): boolean => {
      try {
        // Try to resolve TypeNode to Type... it possible to trigger and exception in some weired cases
        const type = checker.getTypeFromTypeNode(typeNode);

        // Check if TypeNode is resolved to a meaningful Type, not any
        // Also can be validated with (type as unknown as { intrinsicName?: string }).intrinsicName === 'error';
        return type.flags !== ts.TypeFlags.Any;
      } catch (e) {
        return false
      }
    };
    
    return isTypeNodeStrict(checker, typeNode) && !typeArguments.find((node) => !canTypeBeResolved(checker, node));
  }
  ```
  
  Will be easy to read

**Comment by egorio** [comment]
Test case

```
/**
 * The return type is throwing a error
 * @param {[number, number]} ab
 * @param {Array<string>} c
 * @return {Array<{content: string}>}
 */
export function moreCases([a, b], c) {
  a; b; c;
  return [];
}
```

add to packages/codefixes/test/fixtures/base-codefixes/annotateWithStrictTypeFromJSDoc/index.ts

**Review by egorio — CHANGES_REQUESTED** [review]
Test case

```
/**
 * The return type is throwing a error
 * @param {[number, number]} ab
 * @param {Array<string>} c
 * @return {Array<{content: string}>}
 */
export function moreCases([a, b], c) {
  a; b; c;
  return [];
}
```

add to packages/codefixes/test/fixtures/base-codefixes/annotateWithStrictTypeFromJSDoc/index.ts

**Review by wondersloth — COMMENTED** [review]
Looks good. Thank @egorio for improving the solution and adding a test.

**Review by egorio — APPROVED** [review]

File: packages/codefixes/test/__snapshots__/base-codefixes.test.ts.snap [hunk 2]
```diff
@@ -563,6 +563,39 @@ export function think(
...
   return a;
 }
+
+/**
+ * The return type is throwing an error
+ * @param {[number, number]} a-and-b
+ * @param {Array<string>} c
+ * @return {Array<{content: string}>}
+ */
+// @ts-expect-error @rehearsal TODO TS7031: Binding element 'a' implicitly has an 'any' type.
```
egorio, line 573: This works in the next function, but not here.

Created a ticket for this [case](https://github.com/rehearsal-js/rehearsal-js/issues/1062)

Issue #1070, opened 2023-05-19
bug: TS2339 Property 'x' does not exist on type 'Y' caught and stops migration


wondersloth:
Internal Issue working on fixing:
`lib/core/addon/services/linkedin-urls.ts` in `learning-web`

```
$ yarn rehearsal fix --no-graph lib/core/addon/services/linkedin-urls.ts
info:    @rehearsal/fix 3.0.4-beta
✔ Initialize
✖ Debug Failure. False expression: Token end is child end
```

## Test Case

```
import { External } from 'some-package'; // This package has an unknown implementation. No Types.

class Something extends External {
  constructor(...args) {
    super(...args)
    this.setRoutes({
      'enterprise-cookies': "",
      'feed.update': "",
      'help.ads': "",
      'help.binding-benefits': "",
      'help.certificates': "",
      'help.disability-answer-desk': "",
      'help.instructor-analytics': "",
      'help.learning': "",
      'help.learning-privacy': "",
      'help.learning-progress': "",
      'help.learning-teams': "",
      'help.china': "",
      'help.likes': "",
      'help.lms-completion': "",
      'help.lynda-subscriber': "",
      'help.questions-privacy': "",
      'help.reviews': "",
      'help.share-certificate': "",
      'help.sync-learning-history': "",
    });
  }
}
```

## Context

The problem I can deduce with the source file is that it's attempting to call `this.setRoutes()` but it doesn't have that method.

This breaks the internals in rehearsal can throws:

```
Debug Failure. False expression: Token end is child end
```

Running the debugger and iterating over the diagnostics:

We see:

```
"Property 'setRoutes' does not exist on type 'LearningLinkedinUrlsService'."
```

https://github.com/rehearsal-js/rehearsal-js/blame/e06327da9a1f784bbcbda1f9d581ae4ef3652472/packages/codefixes/src/typescript-codefix-collection.ts#L50

### Stack Trace
```
Uncaught Error Error: Debug Failure. False expression: Token end is child end
    at fail (/node_modules/typescript/lib/typescript.js:1811:19)
    at assert (/node_modules/typescript/lib/typescript.js:1830:13)
    at processChildNode (/node_modules/typescript/lib/typescript.js:166208:19)
    at <anonymous> (/node_modules/typescript/lib/typescript.js:166139:11)
    at visitNode2 (/node_modules/typescript/lib/typescript.js:27583:20)
[...]

PR #1071, opened 2023-05-19
fix: getCodeFixesAtPosition fails for TS2339 for specific input with large pojo

wondersloth:
This fixes #1071

This PR fixes an issue when rehearsal attempts to `getCodeFixAtPosition` on a file where the base class can not be determined due to import semantics.

Example Input:

```
import { Base } from 'some-package';

export class Something extends Base {
  constructor() {
    super();
    this.setRoutes({
      'enterprise-cookies': '',
      'feed.update': '',
      'help.ads': '',
      'help.binding-benefits': '',
      'help.certificates': '',
      'help.disability-answer-desk': '',
      'help.instructor-analytics': '',
      'help.learning': '',
      'help.learning-privacy': '',
      'help.learning-progress': '',
      'help.learning-teams': '',
      'help.china': '',
      'help.likes': '',
      'help.lms-completion': '',
      'help.lynda-subscriber': '',
      'help.questions-privacy': '',
      'help.reviews': '',
      'help.share-certificate': '',
      'help.sync-learning-history': '',
    });
  }
}
```
Note: If there is only a couple properties, TypeScript does not error. Though in our use case somewhere around 13 items or more, it starts to freak out.

### Details

- Adds DEBUG to `codefixes` so it will be easier to quickly debug these types of errors (First commit).
- Fixes bug #1070 has reproducible use cases. 
- Creates test for existing error suppression logic.
- Allows for future extension of additional suppressions more easily.

### Why does this have to be fixed?
If we don't suppress this error, any class that extends from a base class from a non-typescript converted package will fail in rehearsal.

### Should we fix this upstream in TypeScript
Yes, but now.

I will be following this up later with a new TS issue.

At this time, I don’t think we should invest time to fix this upstream in TypeScript, as this fix will unblock our internal migrations.


packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -14,12 +15,27 @@ import { Diagnostics } from './diagnosticInformationMap.generated.js';
...
 
+const DEBUG_CALLBACK = debug('rehearsal:codefixes:TypeScriptCodeFixCollection');
+
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 const require = Module.createRequire(import.meta.url);
 
 const { SemicolonPreference, getDefaultFormatCodeSettings } = ts;
 
+type SupressedError = { code: number; message: string };
```
lynchbomb, line 26: `SuppressedError`

packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -14,12 +15,27 @@ import { Diagnostics } from './diagnosticInformationMap.generated.js';
...
+
+const SUPRESSED_ERRORS: Array<SupressedError> = [
```
lynchbomb, line 28: `SUPPRESSED_ERRORS:`

packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -57,13 +73,23 @@ export class TypescriptCodeFixCollection implements CodeFixCollection {
         userPreferences
       );
     } catch (e) {
-      const hideError =
-        diagnostic.code == Diagnostics.TS2345.code &&
-        e instanceof TypeError &&
-        e.message.includes(`Cannot read properties of undefined (reading 'flags')`);
+      const code = diagnostic.code;
+      const supressedFound = SUPRESSED_ERRORS.find((d) => d.code == code);
```
lynchbomb, line 77: suppressedFound

packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -57,13 +73,23 @@ export class TypescriptCodeFixCollection implements CodeFixCollection {
...
+
+      const supressError: boolean =
```
lynchbomb, line 79: suppressError

packages/codefixes/test/__snapshots__/supressed-errors.test.ts.snap
```diff
@@ -0,0 +1,48 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Supressed Errors > TS2339: False expression: Token end is child end, supressed-errors/missingMethod.ts 1`] = `
+"// @ts-expect-error @rehearsal TODO TS2307: Cannot find module 'some-package' or its corresponding type declarations.
+import { Base } from \\"some-package\\";
```
lynchbomb, line 5: wondering in the instance when ts *can* find the module `some-package` are we still seeing this error?

lynchbomb (changes requested):
couple minor things. once tests are passing this should be good

packages/codefixes/test/__snapshots__/supressed-errors.test.ts.snap
wondersloth: In learning-web, I updated a tsconfig to resolve imports to that package, and it still failed. Despite being able to click through the package name to get to the implementation.

This problem is less about the actual class and something breaking given the method not found on the class, and the large size of the pojo.

lynchbomb (approved):

PR #1076, opened 2023-05-22
fix: generate globs only for directories

chadhietala:
In #1067 we regressed as we needed to check to see if we are passing a directory or a file path. A file path is validated downstream.


egorio (approved):

packages/smoke-test/test/commands/__snapshots__/fix.test.ts.snap
```diff
@@ -1,5 +1,230 @@
...
+}
+"
+`;
+
+exports[`fix command > base_ts_app (file only) 4`] = `
+"// based on a given cell, return the number of live neighbors
+export function getLiveNeighborCount(
+  // @ts-expect-error @rehearsal TODO TS7006: Parameter 'grid' implicitly has an 'any' type.
+  grid,
+  coords: number[] | [any, any],
```
chadhietala, line 110: @egorio we seem to be allow tuple types with `any` in them

Issue #1078, opened 2023-05-22
bug: when migrating a test, and running fix twice, source is broken after plugin applied


wondersloth:
Affected Version `@rehearsal/cli@3.0.5-beta`

## Steps to reproduce

Create a `tsconfig.json` in `packages/addons/asset-loader/tsconfig.json`
```json
{
  "extends": "../../../tsconfig.base.json",
  "include": ["addon", "tests", "../../../types/**/*"],
  "references": [],
  "compilerOptions": {
    "allowJs": true,
    "composite": true,
    "declarationDir": "dist-types",
    "paths": {
      "ember-cli-pemberly-tracking/*": [
        "./node_modules/@linkedin/ember-cli-pemberly-tracking/*"
      ],
      "ember-cli-pemberly-tracking/test-support/*": [
        "./node_modules/@linkedin/ember-cli-pemberly-tracking/test-support/*"
      ]
    }
  }
}
```

```bash
$ yarn add -D @rehearsal/cli@3.0.5-beta
$ yarn rehearsal move packages/addons/asset-loader/tests
$ yarn rehearsal fix packages/addons/asset-loader/tests/
// Make a change? Or not.
$ yarn rehearsal fix packages/addons/asset-loader/tests/
```

Second pass of fix breaks source code, due to plugins not acocunting for offset changes when working through fixes.

First pass
```
      assert.ok(
        // @ts-expect-error @rehearsal TODO TS2339: Property 'loadBundleStub' does not exist on type 'TestContext'.
        this.loadBundleStub.calledWithExactly('custom-bundle'),
        'asset-loader service was called with correct bundle name'
      );
```

```
      // @ts-expect-ert'.
      // @ts-expect-error @rehearsal TODO TS2339: Property 'defer' does not exist on type 'TestContext'.
      this.defer.re TODO Property 'loadBundleStub' does not exist on type 'TestContext'.
        // @ts-expect-error @rehearsal TODO TS2339: Property 'this' does not exist on type '"TestContext"'.
        this.loadBundleStub.calledWithExactly('custom-bundle'),
        'asset-loader service was called with correct bundle name'
      );
```

Rehearsal CLI reports things completed.
```
➜  voyager-web git:(medwards/asset-loader) ✗ yarn rehearsal fix packages/addons/asset-loader/tests
info:    @rehearsal/fix 3.0.5-beta
✔ Initialize
✔ Analyzing project dependency graph ...
✔ Types Inferred
[...]

PR #1080, opened 2023-05-22
chore(glint): create test case for multiline hbs in tagged template expression

wondersloth:
Creates a test case for replicating #1078


lynchbomb:
@wondersloth whats the status of this?

wondersloth:
@lynchbomb Need to revalidate. Since we dropped inserting glint comments this may be a moot issue.

wondersloth:
@lynchbomb This is still an issue. When an inline hbs region is found it will parse it incorrectly and start injecting weird types and mangling the file.

PR #1089, opened 2023-05-25
fix: fix service string resolution to support package with org in transform plugin 

### LINKED ISSUE #1087: bug: tests in packages/migrate/tests/migrate.test.ts dont' exercise correct .gts input
## Summary

- The `.gts` tests are not working in `packages/migrate/test/migrate.test.ts`,
- The `.gts` file input is replaced with the '.ts` file contents.
- Input fixtures are nearly identical except for the `<template>` tag.
- Expected, snapshots don’t show the `<template>` tag
- Validated this is the issue, by renaming the class from `SomeComponent` to `SomeGtsComponent` or asserting on that string in the output.

## Details
An example of a bad test:
https://github.com/rehearsal-js/rehearsal-js/blob/f833904932b16909db7f1a18db3695d38e691528/packages/migrate/test/migrate.test.ts#L706

```
    test('when missing a local prop', async () => {
      const [inputs, outputs] = prepareInputFiles(project, ['missing-local-prop.gts']);

      const input: MigrateInput = {
        projectRootDir: project.baseDir,
        packageDir: project.baseDir,
        filesToMigrate: inputs,
        reporter,
      };

      for await (const _ of migrate(input)) {
        // no ops
      }
      expectFile(outputs[0]).contains('class Hello extends');
      expectFile(outputs[0]).toMatchSnapshot();
    });
```

Input
**missing-local-prop.gts**
```
import Component from '@glimmer/component';

export default class Hello extends Component {
  <template>
    <span>Hello, I am {{this.name}} and I am {{@age}} years old!</span>
  </template>
}

```

```
exports[`fix > .gts > when missing a local prop 1`] = `
"import Component from '@glimmer/component';

export default class Foo extends Component {}
"
`;
```

wondersloth:
This fixes #1087.

## Summary

- Prioritizes the `.rehearsal/service-map.json` resolution over our derivation logic.
  - This will allow any future service resolution issues to be short circuited with simply adding a entry to `services-map.json`, rather than waiting for parsing to fail. 
- Improves service strings parsing to support a packageName.
  - This is an internal problem used in some 400+ places in voyager-web. 
```
@service('jet') jet; 
@service('some-module-name@foo') foo;
@service('@some-org/some-package@locale') locale; // This support was added./
```

### DIFF
--- a/packages/migrate/README.md
+++ b/packages/migrate/README.md
@@ -1,3 +1,12 @@
 # @rehearsal/migrate
 
 Rehearsal JavaScript to TypeScript Migration Tool
+
+## Testing
+
+Use the following commands to prepare and clean fixtures for this package:
+
+```
+yarn fixtures:prepare
+yarn fixtures:clean
+```
--- a/packages/migrate/test/__snapshots__/migrate.test.ts.snap
+++ b/packages/migrate/test/__snapshots__/migrate.test.ts.snap
@@ -430,7 +430,9 @@ declare foo: Foo;
 `;
 
 exports[`fix > with addon service > .ts 1`] = `
-"// @ts-expect-error @rehearsal TODO TS2306: File './node_modules/test-addon/services/foo.ts' is not a module.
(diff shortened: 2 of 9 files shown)


egorio (approved):
👍

PR #1095, opened 2023-05-30
refactor: cleanup report summary

lynchbomb:
This PR:

- Fixes a bug where absolute paths in the report key `basePath` | `reportOutDir` was an absolute path.
- Refactors the @rehearsal/reporter report `summary` and removes legacy keys:

```
commandName
entrypoint
basePath
reportOutDir
```

- `commandName`: a given report can only be generated via `rehearsal fix` as such we dont need to explicitly call out which rehearsal command was used to generate the report (only 1 option)
- `entrypoint`: this is a remnant from previous API
- `basePath` | `reportOutDir`: `basePath` was renamed to `reportOutDir`. However,  the underlying value for this path just doesn't provide value. As this was only specifying the path in the report where the report itself resides.

### DIFF
--- a/packages/migrate/test/__snapshots__/migrate.test.ts.snap
+++ b/packages/migrate/test/__snapshots__/migrate.test.ts.snap
@@ -417,18 +417,6 @@ exports[`fix > EXPERIMENTAL_MODES, RehearsalService > mode: single-pass 1`] = `
 "
 `;
 
-exports[`fix > with addon service > .gts 1`] = `
-"import type Foo from 'my-addon/services/foo';
-import Component from '@glimmer/component';
-import { inject as service } from '@ember/service';
-
-export default class SomeComponent extends Component {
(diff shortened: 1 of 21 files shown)


packages/migrate/test/__snapshots__/migrate.test.ts.snap
```diff
@@ -417,18 +417,6 @@ exports[`fix > EXPERIMENTAL_MODES, RehearsalService > mode: single-pass 1`] = `
...
-import Component from '@glimmer/component';
-import { inject as service } from '@ember/service';
-
-export default class SomeComponent extends Component {
-  @service(\\"my-addon@foo\\")
-declare foo: Foo;
-}
-"
-`;
-
```
wondersloth, line 431: Why was this removed?

wondersloth (approved):
Approved. Minor question why a snapshot was removed.

Issue #1101, opened 2023-06-01
move and fix files only under provided target src path


egorio:
The `move` and `fix` uses graph under the hood by default.
The graph's output can contain files that are outside of the target path, for example:

```
src/files1.ts
src/sub/files2.ts // contains `import * from 'file1.ts'`
```
`move src/sub` of `fix src/sub` will modify not only `files2.ts` but also the `files1.ts` because it's used by `files2.ts` and it's part of the graph. 

But when move and fix are called, it's expected to modify files only under provided target path, in the example before it means to modify only `files2.ts` because it's only file under `src/sub`.

So, we need to filter out all files outside of the target path in `move` and `fix` commands.

PR #1102, opened 2023-06-01
fix: move and fix to modify file under the targetPath

egorio:
Filter out files outside of `targetPath` to `move` and `fix` to make commands behavior more predictable. 

Fixes https://github.com/rehearsal-js/rehearsal-js/issues/1101


packages/cli/test/commands/fix/__snapshots__/fix.test.ts.snap
```diff
@@ -42,15 +43,13 @@ These rules, which compare the behavior of the automaton to real life, can be co
   All other live cells die in the next generation. Similarly, all other dead cells stay dead.
 */
 
-// @ts-expect-error @rehearsal TODO TS2307: Cannot find module 'node:process' or its corresponding type declarations.
```
egorio, line 45: This is fixed by https://github.com/rehearsal-js/rehearsal-js/pull/1102/files#diff-9db2b41974dec78bf1531a30673ff621b115da09c4c5c943e606204bca78e381R106

packages/cli/src/commands/fix/tasks/convert-task.ts
```diff
@@ -44,14 +44,18 @@ export function convertTask(
 
       // this just cares about ts files which are already in the proper migration order
       if (orderedFiles) {
+        // Filter out all files that outside of the target directory
+        const filesToMigrate = ctx.orderedFiles.filter((file) =>
+          file.startsWith(resolve(options.rootPath, targetPath))
```
egorio, line 49: I think we need to make `targetPath` absolute in the first place right in cli's `.action` since we do this conversion in a lot of places.

packages/smoke-test/test/commands/__snapshots__/fix.test.ts.snap
```diff
@@ -7,30 +7,27 @@ exports[`fix command > base_ts_app (file only) 1`] = `
...
+[TITLE]   0 have been fixed by rehearsal
+[TITLE]   2 errors need to be fixed manually
+[TITLE]     -- 2 ts errors, marked by @ts-expect-error @rehearsal TODO
 [TITLE]     -- 0 eslint errors, with details in the report
 [SUCCESS] Types Inferred
-[SUCCESS]   12 errors caught by rehearsal
-[SUCCESS]   6 have been fixed by rehearsal
-[SUCCESS]   6 errors need to be fixed manually
-[SUCCESS]     -- 6 ts errors, marked by @ts-expect-error @rehearsal TODO
+[SUCCESS]   2 errors caught by rehearsal
```
lynchbomb, line 18: looks like something is off here. these shouldn't go down.

lynchbomb (changes requested):
looks like a snapshot is regressing with this change

packages/smoke-test/test/commands/__snapshots__/fix.test.ts.snap
egorio: It was originally wrong. This is the correct output because the test is to convert a single file.

lynchbomb:
there is something to say about this workflow which we currently support:
```
rehearsal graph addon-1/src
... migration graph order:
addon-1/src/foo.js
addon-2/src/blah.js
```

then if i subsequently run `move` and `fix` with `--graph` I would expect both of the files above to be moved and fixed. however with this PR change `addon-2` will show in migration graph order however it will NOT be included in `move` and `fix`. essentially the other side of the coin.

lynchbomb:
thoughts on doing:

`--graph` is a boolean. when truthy it will match what the `graph` command is doing. when falsy (default which is not specifying it at all) it will run in this "scoped-graph" mode?

README.md
```diff
@@ -131,11 +131,15 @@ Usage: rehearsal graph [options] [srcPath]
 produces the migration order of files given a source path
 
 Arguments:
-  srcPath                  path to a directory or file (default: current working directory)
+  srcPath                  path to a directory or file (default: process.cwd())
```
wondersloth, line 134: Restore this, I think `saying current working directory` makes sense. Saying process.cwd() implies node knowledge that some people may not have.

packages/cli/src/commands/fix/tasks/convert-task.ts
```diff
@@ -14,14 +14,14 @@ const __dirname = dirname(__filename);
 const workerPath = resolve(__dirname, 'convertWorker.js');
 
 export function convertTask(
-  targetPath: string,
+  srcPath: string,
```
wondersloth, line 17: Why are you removing targetPath?

packages/cli/src/commands/graph/command.ts
```diff
@@ -40,6 +40,7 @@ graphCommand
         rootPath: options.rootPath,
         ignore: options.ignore,
         externals: options.externals,
+        logGraph: true,
```
wondersloth, line 43: NITl something more explicit here
```suggestion
        outputGraphToConsole: true,
```

packages/cli/src/commands/move/command.ts
```diff
@@ -29,7 +29,10 @@ moveCommand
   .alias('mv')
   .description('graph aware git mv from .js -> .ts')
   .argument('[srcPath]', 'path to a directory or file', '')
-  .option('--no-graph', 'opt out of moving the file(s) with the graph')
+  .option(
+    '--graph',
+    'fixing all file(s) within the graph, which might include files outside of the current directory'
```
wondersloth, line 34:
```suggestion
    'fixing all file(s) within the graph, which might include files outside of the srcPath'
```

wondersloth (changes requested):

packages/cli/src/commands/fix/tasks/convert-task.ts
lynchbomb: we arent removing it; just aligning the naming convention to the other tasks

lynchbomb (approved):

wondersloth (approved):

PR #1108, opened 2023-06-06
fix(graph): include hbs files in orderedFiles

wondersloth:
Depends on #1107 

## Why
Glint errors are not being shown in `.hbs` files because they are never passed to the Glint plugins.

## Summary
- `.hbs` files are being omitted from the ordered files past into Migrate.
- The resolver was ignoring `.hbs`
- Stops an `hbs` file from being parsed by the rest of the resolver.
- Includes #1109 
  - Omit hbs from from printers (dot, json, stdout) for graphOrderTask and graphWorker.


packages/migration-graph/src/resolver.ts
```diff
@@ -90,7 +90,13 @@ export class Resolver {
 
     this.graph.addFileToPackage(packageNode, importerAbsPath);
 
+    // We will not attemptt to parse an .hbs
+    if (extname(importerAbsPath) === '.hbs') {
```
lynchbomb, line 94: this makes sense since we need to include the .hbs file in the graph; otherwise the file wont exist for the glint plugin(s). an issue with this is the `rehearsal graph` command will then include this file in the output artifact and log. this is confusing since the .hbs file(s) in question dont have any imports.

lynchbomb (changes requested):
I dont think it makes sense to include the .hbs file(s) in the graph artifact (.json/.dot) nor in the stdout. However this is gonna be tricky to filter them out, and might not be worth it.

We would have to filter out .hbs from the `PackageGraph` before we write the output https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/graph/tasks/graphWorker.ts#L92

And here to filter out from stdout https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/commands/graph/tasks/graphOrderTask.ts#L99

wondersloth:
Addressed @lynchbomb 's concerns merged PR #1109 into this.

This PR now has less snapshot changes, and only has output changes for fix command.

egorio (approved):

lynchbomb:
Looks like the stdout snapshots need to be updated. they are still showing the .hbs

wondersloth:
> Looks like the stdout snapshots need to be updated. they are still showing the .hbs

These snapshots?
https://github.com/rehearsal-js/rehearsal-js/pull/1108/files#diff-66a2593547bd3e3b12a51ac1c4cdee87e3c4e98e89c5573a6b97c2357356dd3d

If these are shown, a file is being fixed in (`convertTask`) that passes files off to migrate.

It's good to show this because we are mutating `.hbs` files with glint-expect-error codes.

lynchbomb (approved):

PR #1109, opened 2023-06-06
feat(graph): omit hbs from printer output

wondersloth:
Depends on #1108

- Omit hbs from from printers (dot, json, stdout)

Note:
- Test coverage is not obvious here. Resolver doesn't have tests for .hbs files these are primarily caught in smoke-test.


packages/cli/src/commands/graph/tasks/graphOrderTask.ts
```diff
@@ -97,6 +97,7 @@ export function graphOrderTask(
 
       if (outputGraphToConsole) {
         task.output = `Graph order for '${srcPath}':\n\n${orderedFiles
+          .filter((filePath) => extname(filePath) !== '.hbs') // Hide .hbs from output
```
wondersloth, line 100: Maybe refactor, this could be a file filter function we pass down through to generateDotLanguage,generateJson.

This way there is less options in the printers.

wondersloth: Nevemind we shouldn't do that, I don't recommend passing down a function because of the indirection for calling `writeOutput`. Which (I think) will not work with passing data to a worker.

egorio (approved):

PR #1137, opened 2023-06-20
fix(migration-graph): expand hasTypes to include typesVersions check

lynchbomb:
This PR:

- includes `typesVersions` as a viable key for package.json types in the migration-graph PackageNode class.

```
"typesVersions": {
    "*": {
      "*": [
        "types/*"
      ]
    }
  }
```


mikesherov:
Does this need a test?

lynchbomb:
yeah prob. ill add something here. the implementation is checking against the existence of a key in valid package.json

PR #1141, opened 2023-07-01
feat: GlintCodeFix Add Component Signature Interface to `.gts` component class

wondersloth:
# Summary

- Creates a GlintCodeFIx for adding missing arguments from a `.gts` template to the Component Signature.
- This feature will:
  - Add a missing argument to a signature.
  - Add the `Args` property to a signature.
  - Create a signature based off of the component className.
  - Create a signature with the name defined by it's jsDoc comment.
  - Update a signature with a missing arg.

# Details

## How does the GlintCodeFix Work

- Given a specific glint diagnostic `Glint2339/TS2339 Property 'name' does not exist on type 'User'`
- We determine if this within a template `@argument` from the diagnostic.
- From that we inspect if the **component** has a **signature interface** defined.
- If no component signature interface exists, we create one inferring the name from the **component** class.
- If a comment exists for the component signature interface, but it's not bound to the heritage clause, it will read from the comment and update the clause. 
  - This addresses a gap in typescript's codefix for a `@typedef` and extend
- If the Component Signature Interface exists, but no Args property, then create the property.
- If the Component Signature Interface exists and the Args property exits, then add the argument to Args. #1139

## Edge Cases, Not addressed

- Does not work in a `.ts`, only `.gts`.
- Does not work yet with template only components. This will have to be a follow-up.
- Does not attempt to use the component signature in a component's constructor.

### **swc** span positions are byteBuffer

`@swc/core` does not provide character positions for it's `node.span` properties on it's AST nodes. They are actually a position in the buffer.
See: https://github.com/swc-project/swc/issues/1366

This issue suggests (2) work arounds:

1. We calculate the offset with an empty `parseSync` call and store that to get a references to the correct buffer start position.
2. We map the position in the buffer back to the char position. This is handled in `glint-parsing-utils` with `SWCParsedResult` and `StringAsBytes`

This is non-obvious at first. Small usecases work, but glint transformedCode (intermediate files with greek characters `𝛄,χ,𝚪`) have different sizes, which starts to throw off the offset. Hence why we needed solution (2).

```
  console.log(byteSize('𝛄χ𝚪')); // 10
  console.log(byteSize('123')); // 3
```

Also, you don't see these byte positions issues on the first pass of a file in isolation, you see it in long running processes like watching tests.

### Drop swc for ts.

We refactor all the `swc` code in `glint-parser-utils` to use `ts.createSourceFile` instead.

Pros:

- Aligns all CodeFix changes to other rehearsal CodeFixes and Glint internals.

Cons:

- We have to complete this rewrite.

I have this conversion partially started, but I currently have a working solution for swc.

In general I do not recommend **swc** usage if you need to get `span` `start` and `end` positions for
source manipulation. Current tooling works but makes it harder to debug.

I completed the solution with **swc** because I was about 80% done, and got it across the line. I think rewriting to use TS is best.

## Caveats

For best results where the interface is populated completely we need `--mode=drain`.

### Why do we have to use `mode: drain` to get the complete interface?

Typically in the first pass if no component signature interface exits, we will create it.

We currently cannot apply all fixes, because of how are changes are queued and deduped with magic string, prior to updating the source file in glint.

For us to be able to apply multiple fixes, we would have to eagerly write back to source.

# TODO
- [x] Discuss the usage of drain to completely create the Component Signature.
- [x] Nice to Have: Rewrite parsing to drop swc in favor of ts, in the CodeFix. Not happy with the the amount of work that took to get it to play nicely with offsets.

### TBD Glint bug, when environments `ember-loose` and `ember-template-imports` are enabled together

**Update** This is no longer blocking our tests. Unsure the root cause.

In testing, I've found the intermediate transformed source file from glint may occasionally contain both converted templates for `ember-loose` and `ember-template-imports` environment.

See: https://gist.github.com/wondersloth/921979d4ea96695edf4db6b074460faa#file-example-test-ts

The problem here is that the rewritten module has both static method transforms, and causes a problem when mapping back to the original source file due to an out of bounds condition.

For now, in the `packages/migrate/test/migrate.test.ts` fixture, I've removed `ember-loose`.
- [ ] Investigate `ember-loose` and `ember-template-imports` environment in transformed source file.
- [ ] Investigate why with-errors.hbs has no errors reported to reporter. See `migrate.test.ts`


wondersloth:
I've dropped usage of SWC in the CodeFix, because of the ByteBuffer issue. I didn't like how the Buffer Ranges from swc didn't match up when I would manually transform the files and compare offsets. This ultimately lead me to rewrite all parsing utils in this PR to use ts.

packages/codefixes/test/__snapshots__/glint-codefixes.test.ts.snap
```diff
@@ -0,0 +1,55 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Test glint codefixes > addMissingArgToComponentSignature 1`] = `
+"import Component from \\"@glimmer/component\\";
+
+export interface WithMissingArgsSignature {
+  Args: { age: any; snack: any };
```
lynchbomb, line 7: why are we inferring `any` here?

packages/codefixes/test/base-codefixes.test.ts
```diff
@@ -15,8 +15,10 @@ describe('Test base codefixes', function () {
   const codefixesDir = resolve(__dirname, '../src/fixes/');
 
   // eg. ["addErrorTypeGuard","addMissingExport","addMissingTypesBasedOnInheritance","makeMemberOptional"]
-  const transforms = readdirSync(codefixesDir).map((file) => parse(file).name);
-
+  const transforms = readdirSync(codefixesDir)
+    .map((file) => parse(file).name)
+    .filter((item) => !item.startsWith('glint')); // Exclude glint directory from test, this will be tested by glint-codefixes
+  console.log(transforms);
```
lynchbomb, line 21: remove the console.log

packages/codefixes/test/glint-codefixes.test.ts
```diff
@@ -0,0 +1,142 @@
...
+        'ember-template-imports',
+      ],
+      checkStandaloneTemplates: true,
+    };
+
+    project.files['tsconfig.json'] = JSON5.stringify(tsConfig, { space: 2, quote: '"' });
+
+    await project.write();
+
+    execSync(`pnpm install`, {
```
lynchbomb, line 75: this slower test seems more appropriate to be in the smoke-test (adding deps, updating tsconfig and prettier) where we are already doing things like this (pnpm install) and can be run in a separate execution. lets move this entire file into smoke-test.

packages/migrate/test/__snapshots__/migrate.test.ts.snap
```diff
@@ -1,10 +1,82 @@
...
+  </template>
+}
+"
+`;
+
+exports[`fix > .gts > component signature codefix > adds missing args to Args to component signature interface 1`] = `
+"import Component from \\"@glimmer/component\\";
+
+export interface WithMissingArgsSignature {
+  Args: { age: any; snack: any };
```
lynchbomb, line 43: same as prior. why are we inferring `any` here (we should try and infer a type for `age` and `snack`

packages/smoke-test/test/commands/__snapshots__/fix.test.ts.snap
```diff
@@ -679,7 +683,8 @@ import GreetingHeader from \\"./GreetingHeader\\";
  * @param {string} name the name to format
  * @returns {string} the formatted name
  */
-function formatName(name: string): string {
+// @ts-expect-error @rehearsal TODO TS80004: JSDoc types may be moved to TypeScript types.
+function formatName(name: string) {
```
lynchbomb, line 687: looks like we've lost the return type here. should be `function formatName(name: string): string`

lynchbomb (changes requested):
Left a bunch of comment. Looking good overall. If we can infer types for the Component Signature Interfaces that'll be great.

packages/plugins/src/plugins/glint-fix.plugin.ts
```diff
@@ -135,15 +168,21 @@ export class GlintFixPlugin extends Plugin<GlintFixPluginOptions> {
...
+          if (this.hasAppliedChange(targetFileName, change)) {
+            continue; // Skip if a duplicate change
           } else {
-            // this.appliedAtOffset[fileTextChange.fileName].push(change.span.start);
+            // Init
+            if (!this.appliedTextChanges[targetFileName]) {
+              this.appliedTextChanges[targetFileName] = [];
+            }
+            // Append and normalize
+            this.appliedTextChanges[targetFileName].push(change);
```
egorio, line 182: this.appliedTextChanges[targetFileName].push(normalizeTextChanges(change)); ?

packages/plugins/src/plugins/glint-fix.plugin.ts
```diff
@@ -135,15 +168,21 @@ export class GlintFixPlugin extends Plugin<GlintFixPluginOptions> {
...
-          if (this.appliedAtOffset[fileTextChange.fileName].includes(change.span.start)) {
-            continue;
+          // Track the applied changes so we can dedupe as we go
+          if (this.hasAppliedChange(targetFileName, change)) {
+            continue; // Skip if a duplicate change
           } else {
-            // this.appliedAtOffset[fileTextChange.fileName].push(change.span.start);
+            // Init
+            if (!this.appliedTextChanges[targetFileName]) {
+              this.appliedTextChanges[targetFileName] = [];
```
egorio, line 179: We have something new [for this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment):

```
this.appliedTextChanges[targetFileName] ??= []
```

packages/plugins/src/plugins/glint-fix.plugin.ts
```diff
@@ -135,15 +168,21 @@ export class GlintFixPlugin extends Plugin<GlintFixPluginOptions> {
...
-
-          if (this.appliedAtOffset[fileTextChange.fileName].includes(change.span.start)) {
-            continue;
+          // Track the applied changes so we can dedupe as we go
+          if (this.hasAppliedChange(targetFileName, change)) {
+            continue; // Skip if a duplicate change
           } else {
-            // this.appliedAtOffset[fileTextChange.fileName].push(change.span.start);
+            // Init
+            if (!this.appliedTextChanges[targetFileName]) {
```
egorio, line 178: We have something new [for this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment):

```
this.appliedTextChanges[targetFileName] ??= []
```

packages/codefixes/src/fixes/glint/addMissingArgToComponentSignature.ts
```diff
@@ -0,0 +1,218 @@
...
+
+export class AddMissingArgToComponentSignature implements CodeFix {
+  getErrorCodes = (): number[] => [Diagnostics.TS2339.code];
+
+  getCodeAction(diagnostic: DiagnosticWithContext): ts.CodeFixAction | undefined {
+    const glintDiagnostic = diagnostic as GlintDiagnosticWithContext;
+    const glintService = glintDiagnostic.glintService;
+
+    // Only support .gts files and not .hbs
+    if (extname(glintDiagnostic.file.fileName) !== '.gts') {
```
egorio, line 31: Can this be applied to .ts as well?

packages/plugins/src/plugins/glint-fix.plugin.ts
lynchbomb: yeah this is pretty awesome. you no longer have to wrap in a try/catch for missing keys

packages/codefixes/test/__snapshots__/glint-codefixes.test.ts.snap
```diff
@@ -0,0 +1,55 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Test glint codefixes > addMissingArgToComponentSignature 1`] = `
+"import Component from \\"@glimmer/component\\";
```
wondersloth: This should be removed.

egorio (approved):
Great!

packages/codefixes/package.json
```diff
@@ -38,6 +38,7 @@
     "version": "pnpm version"
   },
   "dependencies": {
+    "@babel/parser": "^7.22.5",
```
lynchbomb, line 41: can this be removed? I dont see the import anywhere

lynchbomb (changes requested):
@babel/parser?

wondersloth:
I must have added it by accident when evaluating alternatives to swc given the strong position bug I had. Will remove.

packages/smoke-test/test/commands/__snapshots__/fix.test.ts.snap
wondersloth: This has been addressed.

wondersloth:
Updated the PR addressing all issues. Regressed in our target internal early adopter. Found some issues with `.gts` files getting unsafe changes (unrelated to this PR), created an issue related to that.

lynchbomb (approved):

Issue #1175, opened 2023-07-14
[Feature Request] Add  flags to rehearsal fix that will allow the bypassing of dependency checks.


robostheimer:
It is possible that a consuming app could have a novel setup in which they manage the packages necessary for rehearsal fix outside of the app's package.json (i.e. in another package so that it can be shared across an entire app ecosystem), meaning the required packages may not be in the package.json file but are available in an apps lock file.

The suggestion is to add two flags that will allow a consuming app to bypass the package check  and eslint config checks.

PR #1178, opened 2023-07-17
Adding --skipDepsCheck and --skipEsLintCheck flags to fix command (Issue 1175)

robostheimer:
## Summary
Fix for [Issue 1175](https://github.com/rehearsal-js/rehearsal-js/issues/1175): Add flags to rehearsal fix that will allow the bypassing of dependency checks. 

After consultation with @lynchbomb @wondersloth, it was decided that to fix this issue we could add flags that would allow for the bypass of the dependencies and eslint config checks that run when calling `rehearsal fix`.

New flags can be called individually:
`rehearsal fix --skipDepsCheck`
`rehearsal fix -- skipEsLintCheck`

or together:
`rehearsal fix --skipDepsCheck --skipEsLintCheck`


packages/cli/src/commands/fix/tasks/initialize-task.ts
```diff
@@ -45,7 +45,7 @@ export function initTask(
         projectType = 'glimmer';
       }
 
-      preFlightCheck(rootPath, projectType);
+      preFlightCheck(rootPath, projectType, skipDepsCheck, skipEsLintCheck);
```
lynchbomb, line 48: lets change the last param to a record if we need to expand to include additional `skips`

```
const skipChecks = {
skipDepsCheck,
skipEsLintCheck
}
```

eg. `preFlightCheck(rootPath, projectType, skipChecks)`

packages/cli/src/commands/fix/tasks/initialize-task.ts
```diff
@@ -66,6 +66,8 @@ export function initTask(
 export function preFlightCheck(
   basePath: string,
   projectType: ProjectType,
+  skipDepsCheck = false,
```
lynchbomb, line 69: same here update to take a `skipChecks` record

packages/cli/src/commands/fix/tasks/initialize-task.ts
robostheimer: sure thing

packages/cli/src/commands/fix/command.ts
```diff
@@ -47,6 +47,8 @@ fixCommand
     parseCommaSeparatedList,
     ['sarif']
   )
+  .option('--skipDepsCheck', 'skips the isDepsPreReq check when initializing fix task')
```
lynchbomb, line 50: instead of having to list every skip individually just allow for an array of specified strings of the type of skip

eg. `.option('--skipChecks [checks...]', 'skips the specified checks when initializing fix task eg. eslint, deps, tsconfig', parseCommaSeparatedList)`

lynchbomb (changes requested):
left a few comments

lynchbomb:
@robostheimer  You'll need to rebase the GH Actions will run the test suites

packages/cli/src/commands/fix/command.ts
robostheimer: updated.

packages/cli/src/commands/fix/command.ts
```diff
@@ -47,6 +47,12 @@ fixCommand
     parseCommaSeparatedList,
     ['sarif']
   )
+  .option(
```
lynchbomb, line 50: lets trim this down and expose the list of available skips into an enum as the single source of truth. IIRC commander doesn't support choices with multiple selections, so this is the best we can do.

```
// ... types.ts

// list of all the allowed checks which can be skipped
export enum SKIP_CHECKS {
  eslint = 'eslint',
  deps = 'deps',
}

// create the type from the enum
export type SkipChecks = (keyof typeof SKIP_CHECKS)[] | undefined;

// ... command.ts

import { SKIP_CHECKS } from '../types'

// generate the allowed checks in the help command from the enum
`skip specified pre-flights checks (not recommended), eg. --skipChecks ${Object.keys(SKIP_CHECKS).join(', ')}`
```

README.md
```diff
@@ -182,12 +184,15 @@ Arguments:
...
-                       tests/**/*,types/**/*' (default: [])
-  -d, --dryRun         do nothing; only show what would happen (default: false)
-  -h, --help           display help for command
+  --graph                   fixing all file(s) within the graph, which might include files
+                            outside of the current directory
+  --ignore [globs...]       comma-delimited list of globs to ignore eg. '--ignore
+                            tests/**/*,types/**/*' (default: [])
+  -d, --dryRun              do nothing; only show what would happen (default: false)
+  -h, --help                display help for command'
+  --skipChecks [checks...]  skips check for packages, useful if a codebase has a novel
```
lynchbomb, line 193: once the change in the command itself has been made just update this to match

packages/cli/src/types.ts
```diff
@@ -59,8 +59,11 @@ export type FixCommandOptions = {
   ignore: string[];
   graph?: boolean;
   mode: 'single-pass' | 'drain';
+  skipChecks?: SkipChecks;
 };
 
+export type SkipChecks = ('eslint' | 'deps')[];
```
lynchbomb, line 65: // list of all the allowed checks which can be skipped
export enum SKIP_CHECKS {
  eslint = 'eslint',
  deps = 'deps',
}

// create the type from the enum
export type SkipChecks = (keyof typeof SKIP_CHECKS)[] | undefined;

lynchbomb (changes requested):
Coming along. I think you might have rebased commits from outside of this PR. Just a note when rebasing be sure to:

```
  git checkout master
  git pull --rebase
  git checkout YOUR_BRANCH
  git rebase master
```

README.md
robostheimer: Done

packages/cli/src/commands/fix/command.ts
robostheimer: Done

packages/cli/src/types.ts
robostheimer: Done

robostheimer:
> Coming along. I think you might have rebased commits from outside of this PR. Just a note when rebasing be sure to:
> 
> ```
>   git checkout master
>   git pull --rebase
>   git checkout YOUR_BRANCH
>   git rebase master
> ```

I was able to successfully remove the offending commits.  Note rebasing is not quite as simple as described here, because this is a fork of the branch.  Actually appears that rebasing should occur in the github UI using the `Sync fork` button, which is much easier than what I did 🤷 .
Screenshot 2023-07-19 at 10 14 19 AM

lynchbomb (approved):

Issue #1192, opened 2023-07-31
Bug: Second pass of rehearsal fix, breaks when jsdoc inferred type has invalid name


wondersloth:
## Use Case
* When a JsDoc comment has invalid characters for a variable name e.g. `--`
* When run fix multiple times against a file: `fix(fix(content))`

## Bug from Internal Code

```
  /**
   * This action is yielded to the consumer to allow them to
   * set/reset the keywords based on some user action. The primary
   * use case intended is for an input clear button.
   *
   * @param {Event} -- Optional DOM event passed when action is bound to event element attribute
   * @param {string} -- Optional value to reset keywords to
   * @param {*} -- Optional value to reset selection state to
   */
  someFunc(...args) {
    //
  }
```

## Reproduction
Given a `.gts` using a JsDoc Syntax `--` for an unknown argument type and name

## Input

```
export default class JsDocErrorsExample {
  /**
   // @ts-expect-error @rehearsal TODO TS8024: JSDoc '@param' tag has name '', but there is no parameter with that name.
   * @param {Event} -- Something
   // @ts-expect-error @rehearsal TODO TS8024: JSDoc '@param' tag has name '', but there is no parameter with that name.
   * @param {string} -- Another
   // @ts-expect-error @rehearsal TODO TS8024: JSDoc '@param' tag has name '', but there is no parameter with that name.
   * @param {*} -- I don't know
   */
  reset(...args) {
    console.log(args);
  }

  shouldExist() {
    console.log("should exist in output");
  }
}
```

## Output

```
export default class JsDocErrorsExample {
  // @ts-expect-error @rehearsal TODO TS1005: '}' expected.
  

 
```

I was unable to write a `.ts` file that would inject those rehearsal TODOs that would cause this error.
```
// @ts-expect-error @rehearsal TODO TS8024: JSDoc '@param' tag has name '', but there is no parameter with that name.
```

I was how ever able to make a `.gts` use case that would cause this, replicating the bug I found internally

#### Input
```
import Component from "@glimmer/component";

export default class JsDocErrorsExample extends Component {
  /**
   * @param {Event} -- An Event
   * @param {string} -- Some String
   * @param {*} -- Something unknown
   */
  reset(...args) {
    console.log(args);
  }

  shouldExist() {
    console.log("should exist in output");
  }

  <template>
    Hello World
  </template>
}

```

#### Output from `first-pass`
```
import Component from "@glimmer/component";

[...]

PR #1193, opened 2023-08-01
feat: stubMissingNameJSDocParam

wondersloth:
This PR addresses #1192 

## Summary

There is a use case in an internal app where a JSDoc `@param` is defined but has no variable name. This creates a case where we inject a rehearsal TODO into the doc block, and subsequent invocations of `rehearsal fix` will mangle the file after the affected region.

## Root Cause

Transforming a doc block with this will cause a second pass with fix to mangle the output of the file there after.

```diff
  /**
+   // @ts-expect-error @rehearsal TODO TS8024: JSDoc '@param' tag has name '', but there is no parameter with that name.
   * @param {Event} -- Something
+   // @ts-expect-error @rehearsal TODO TS8024: JSDoc '@param' tag has name '', but there is no parameter with that name.
   * @param {string} -- Another
+   // @ts-expect-error @rehearsal TODO TS8024: JSDoc '@param' tag has name '', but there is no parameter with that name.
   * @param {*} -- I don't know
   */
```

## Details

The type error we have is `TS8024`, is for a JSDoc Param name that doesn’t appear in the scope of the function.

Use Case

```typescript
  /**
   * This action is yielded to the consumer to allow them to
   * set/reset the keywords based on some user action. The primary
   * use case intended is for an input clear button.
   *
   * @param {Event} -- Optional DOM event passed when action is bound to event element attribute
   * @param {string} -- Optional value to reset keywords to
   * @param {*} -- Optional value to reset selection state to
   */
  @action
  reset(...args) {
    const firstArg = args[0];
    let resetArgs;

    // the first argument may be an event if
    // the action is bound to an event attribute
    // if that's the case we don't want to include
    // the event in the args passed to `this.reset`
    if (firstArg instanceof window.Event) {
      resetArgs = args.slice(1);
    } else {
      resetArgs = args;
    }

    this._reset(...resetArgs);
  }
```

```typescript
import Component from '@glimmer/component';

export class Something extends Component {
  /**
   * @param {Event} -- Something
   * @param {string} -- Another
   * @param {*} -- I don't know
   */
  reset(...args) {
    console.log(args);
  }

  shouldExist() {
    console.log('should exist in output');
  }
}
```

```typescript
import Component from '@glimmer/component';

export class Something extends Component {
  /**
   // @ts-expect-error @rehearsal TODO TS8024: JSDoc '@param' tag has name '', but there is no parameter with that name.
   * @param {Event} -- Something
   // @ts-expect-error @rehearsal TODO TS8024: JSDoc '@param' tag has name '', but there is no parameter with that name.
   * @param {string} -- Another
   // @ts-expect-error @rehearsal TODO TS8024: JSDoc '@param' tag has name '', but there is no parameter with that name.
   * @param {*} -- I don't know
   */
  // @ts-expect-error @rehearsal TODO TS7019: Rest parameter 'args' implicitly has an 'any[]' type.
  reset(...args) {
    console.log(args);
  }

  shouldExist() {
    console.log('should exist in output');
  }
}
```

When ever a rehearsal todo like above is injected into a doc block subsequent applications of `fix` mangle the file to look like:

```typescript
import Component from "@glimmer/component";

export class Something extends Component {
```

## Solution

Stubbing the missing parameter name will trigger the unused argument codefix as it’s able to check for the value in scope and remove the unused param.

With stubbing the argument to something like `@param Event unnamedParam_sha123`

### Adding `eslint-plugin-jsdoc` can help with general JSDoc errors

This works for some errors, but not in our use case.

https://github.com/gajus/eslint-plugin-jsdoc#readme

For something like:

```
  /**
   * @param {Event} -- Something
   * @param {string} -- Another
   * @param {*} -- I don't know
   */
```

eslint naively sees `--` as the param name, and sees those entries as duplicates.

```
   4:1  warning  Duplicate @param "--"     jsdoc/check-param-names
```

### Use existing AnnotateWithStrictTypeFromJSDoc CodeFix to `.gts` files

That CodeFix does apply to the error we are trying to suppress.

### DIFF
--- a/packages/codefixes/src/fixes/stubMissingJSDocParamName.ts
+++ b/packages/codefixes/src/fixes/stubMissingJSDocParamName.ts
@@ -6,15 +6,14 @@ import { Diagnostics } from '../diagnosticInformationMap.generated.js';
 import type { CodeFix, DiagnosticWithContext } from '../types.js';
 
 function createUniqueIdentifier(data: string): string {
-  const hash = createHash('md5').update(data).digest('base64');
+  const hash = createHash('md5').update(data).digest('hex');
   return hash.substring(0, 6);
 }
 
-export class StubMissingNameJSDocParam implements CodeFix {
+export class StubMissingJSDocParamName implements CodeFix {
   getErrorCodes = (): number[] => [Diagnostics.TS8024.code];
 
   getCodeAction(diagnostic: DiagnosticWithContext): CodeFixAction | undefined {
-    const fileName = diagnostic.file.fileName;
     const originalStart = diagnostic.start;
     const originalEnd = originalStart + diagnostic.length;
     const content = diagnostic.file.getFullText(diagnostic.file);
@@ -23,13 +22,15 @@ export class StubMissingNameJSDocParam implements CodeFix {
       return;
     }
 
-    const uniqueIdentifier = createUniqueIdentifier(
-      `${fileName}-${content}-${originalStart}-${originalEnd}`
-    );
+    const uniqueIdentifier = createUniqueIdentifier(`${content}-${originalStart}-${originalEnd}`);
 
     const changes: [ts.FileTextChanges] = [
       // Augment the SuperClass signature first as it's lower in the file
-      ChangesFactory.insertText(diagnostic.file, originalStart, `unnamedParam${uniqueIdentifier}`),
+      ChangesFactory.insertText(
+        diagnostic.file,
+        originalStart,
+        `unnamedParam_${uniqueIdentifier} `
+      ),
     ];
 
     return createCodeFixAction(
--- a/packages/codefixes/src/fixes/stubMissingJSDocParamName.ts
+++ b/packages/codefixes/src/fixes/stubMissingJSDocParamName.ts
@@ -36,7 +36,7 @@ export class StubMissingJSDocParamName implements CodeFix {
     return createCodeFixAction(
       'stubMissingNameJSDocParam',
       changes,
-      'adds a unique name to the unamed JSDoc @param'
+      'adds a unique name to the unnamed JSDoc @param'
     );
   }
 }
--- a/packages/migrate/test/__snapshots__/migrate.test.ts.snap
+++ b/packages/migrate/test/__snapshots__/migrate.test.ts.snap
@@ -111,6 +111,84 @@ export default class WithMissingInterface extends Component<WithMissingInterface
 "
 `;
 
+exports[`fix > .gts > component signature codefix > with misssing jsdoc param name 1`] = `
+"import Component from \\"@glimmer/component\\";
+
+export class Something extends Component {
+  /** */
+  // @ts-expect-error @rehearsal TODO TS7019: Rest parameter 'args' implicitly has an 'any[]' type.
+  reset(...args) {
+    console.log(args);
+  }
+
+  shouldExist() {
+    console.log(\\"should exist in output\\");
+  }
+}
+"
+`;
+
+exports[`fix > .gts > component signature codefix > with typedef for component signature interface 1`] = `
+"import Component from \\"@glimmer/component\\";
+
+export interface RepeatSignature {
+  Args: { phrase: any };
+}
+
+//
+// UseCase: Should generate a component signature interface
+//
+
+export class Repeat extends Component<RepeatSignature> {
+  <template>
+    <span>{{@phrase}}</span>
+  </template>
+}
+
+//
+// UseCase: Should infer relationship from comment, update heritage clause with signature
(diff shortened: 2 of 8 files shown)


packages/codefixes/src/fixes/stubMissingJSDocParamName.ts
```diff
@@ -0,0 +1,42 @@
...
+        diagnostic.file,
+        originalStart,
+        `unnamedParam_${uniqueIdentifier} `
+      ),
+    ];
+
+    return createCodeFixAction(
+      'stubMissingNameJSDocParam',
+      changes,
+      'adds a unique name to the unamed JSDoc @param'
```
wondersloth, line 39:
```suggestion
      'adds a unique name to the unnamed JSDoc @param'
```

wondersloth:
Validated this in voy-web with the second-pass not resulting in a syntax error or mangled file.

packages/migrate/test/__snapshots__/migrate.test.ts.snap
```diff
@@ -185,6 +263,41 @@ export default class TestGjsNoErrors extends Component {
 "
 `;
 
+exports[`fix > .gts > strips jsdoc param with missing name 1`] = `
+"import Component from \\"@glimmer/component\\";
+
+export class Something extends Component {
+  /** */
```
egorio, line 270: This is happening because of `deleteUnmatchedParameter`

packages/codefixes/src/fixes/stubMissingJSDocParamName.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    const uniqueIdentifier = createUniqueIdentifier(`${content}-${originalStart}-${originalEnd}`);
+
+    const changes: [ts.FileTextChanges] = [
+      // Augment the SuperClass signature first as it's lower in the file
+      ChangesFactory.insertText(
+        diagnostic.file,
+        originalStart,
+        `unnamedParam_${uniqueIdentifier} `
```
egorio, line 32: I don't think you need `uniqueIdentifier`, I think you can just keep `unnamedParam`.

egorio: I would actually add param from function in order, if their number are matched

```
/**
   * @param {Event} -- Something
   * @param {string} -- Another
   * @param {*} -- I don't know
   */
  reset(a, b, c): void {
```
to
```
/**
   * @param {Event} a -- Something
   * @param {string} b -- Another
   * @param {*} c -- I don't know
   */
  reset(a, b, c): void {
```
this will help to infer types and will prevent tags to be delated.
  
  
In cases like this (one destructed param), [you can use](https://jsdoc.app/tags-param.html#multiple-types-and-repeatable-parameters) [something like this](https://www.typescriptlang.org/play?ssl=10&ssc=2&pln=1&pc=1#code/PQKhCgAJpSAEAOBDATkgtpA3gUQG4CmAdgC4C+kqA5gM4DaAugHQAMkAtO5AMoD26BEgAsAlkSpQY8ZGkxYaJFGKoVq9ZgEYOXAIJFewgikmRYiVBmwhVKWoyYAmbZACSkACa8iAchKQA1voA7iYgwOAAZgCuRADGJCJekCgENIIAFExZagCUAFyQeLwi7tiSsV4KlJAAvJS29CwMANzllX4ARrX1dhotbURVsd1qdA79ZEA):
```
/**
   * @param {Event} args[].0 -- Something
   * @param {string} args[].1 -- Another
   * @param {*} args[].2 -- I don't know
   */
  reset(...args): void {
```

egorio (approved):

egorio:
Please read https://github.com/rehearsal-js/rehearsal-js/pull/1193#discussion_r1281164865 before merge.

packages/codefixes/src/fixes/stubMissingJSDocParamName.ts
wondersloth: Talked with @egorio over slack.

We can improve this code fix by inferring types a later date.

This CodeFix specifically addresses a JSDoc syntax error when a variable name is not defined.

If we do improve on this we have to be guarded about how much we want to infer from the source and what is the correct type and positions. For now this CodeFix addresses our needs but could be improved to improve JsDoc and thus types.

packages/migrate/test/__snapshots__/migrate.test.ts.snap
wondersloth: Ah, yes. I saw that in our tests but didn't know what they were for. We have tests covering ts-codefixes so we can validate the changes are operating as we expect them to.

Issue #1187, opened 2023-07-25
adds type to signature, however does not import the type


lynchbomb:
# Build
Latest `1.1.0`

# Error
This throws the CLI and exits:
```
An unknown error occured when attemping to getCodeFixesAtPosition for file: voyager-web/packages/addons/abi-shared/addon/routes/abi/results.ts due to TS7006 Parameter 'controller' implicitly has an 'any' type
```

# Replicate
`WORKER=false yarn rehearsal fix voyager-web/packages/addons/abi-shared/addon/routes/abi/results.ts`

```
import type Controller from '@ember/controller'; <<< missing. rehearsal is not importing the type which it has inferred (below)

  override resetController(
    controller: Controller, <<< rehearsal is inferring the "Controller" type to the signature
    isExiting: boolean,
```

# Resolve
If you simply add the missing type import to the file above `import type Controller from '@ember/controller';`. And then re-run rehearsal everything works as expected.

Whats curious is this is the ONLY type `Controller` which is causing this issue. Rehearsal can correctly infer and import other missing types. Additionally this happens on every file where this `Controller` type is inferred and missing the import.

lynchbomb:
@egorio did some investigation here:

The failure is coming from TS `inferFromUsage` codefix on `setupController`. You can't see this suggestion in IDE because codefix is failing. Not directly related to imported / not imported Controller. I'm looking for fixes for both import of Controller and avoiding the failure above. 

For "import issue":

- quick fix is to prioritize TS2304 (one of codefixes is to fix missing import)
- the right fix is to improve "addMissingTypeBasedOnInheritance" by importing same type as in the file containing the parent class / interface

PR #1215, opened 2023-08-02
fix: import types when infer from inheritance

egorio:
Prioritize "Fixes missing imports" error codes to make sure we don't have missing types before start types inferring process.

This helps to solve the issue when type are inferred from inheritance or from inlay type hints without been imported. The type will be imported next cycle (in drain mode) and the rest of codefixes expected to be applied without issues. 

Also fixes: 
- module resolutions on case-insensitive filesystems

Fixes https://github.com/rehearsal-js/rehearsal-js/issues/1187

### DIFF
--- a/packages/migrate/test/migrate.test.ts
+++ b/packages/migrate/test/migrate.test.ts
@@ -1207,6 +1207,7 @@ export default class LocaleService extends Service {
         packageDir: project.baseDir,
         filesToMigrate: inputs,
         reporter,
+        mode: 'drain',
       };
 
       for await (const _ of migrate(input)) {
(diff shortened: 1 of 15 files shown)


packages/service/src/rehearsal-service-host.ts
```diff
@@ -32,9 +33,17 @@ export class RehearsalServiceHost implements LanguageServiceHost {
     this.fileNames = fileNames;
   }
 
+  useCaseSensitiveFileNames = (): boolean => {
```
egorio, line 36: Compiler resolves paths based on this function result :)

wondersloth: This was a weird issue.

packages/migrate/test/migrate.test.ts
```diff
@@ -1207,13 +1207,38 @@ export default class LocaleService extends Service {
...
+        filesToMigrate: inputs,
+        reporter,
+        mode: 'drain',
       };
 
       for await (const _ of migrate(input)) {
         // no ops
       }
 
       expectFile(outputs[0]).toMatchSnapshot();
```
wondersloth, line 1239: Add an assertion for the file that should have a type e.g. `feed(food: Food)` in the dog file assertion.

This will help in situations that we have snapshots updated without validating output.

wondersloth (changes requested):

lynchbomb (approved):

packages/migrate/test/migrate.test.ts
lynchbomb: @egorio once this is resolved and validated in our consumer apps, this is good to merge

egorio: Added. Will merge PR after validation.

wondersloth (approved):

Issue #1220, opened 2023-08-03
bug: Glint Service is not inferring return types from inlay hints 


egorio:
This is happening because Glint Service is missing [getAdditionalDiagnostics](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/service/src/rehearsal-service.ts#L113C3-L137) function. 

Need to "copy" this function from Rehearsal Service.

PR #1222, opened 2023-08-04
fix: Glint Service is not inferring return types from inlay hints

egorio:
Adds a `getAdditionalDiagnostics` function to Glint Service to infer return types based on inlay hints.

Also:
- Starting now, `.gts` files have `Method 'containsUnsupportedDiagnostic' lacks a return-type annotation.` comments over function with missing return type.
- `.ts` under Glint projects will have return types inferred where type hints are available..
- SourceFiles produced by Glint Service for .gts now compatible with TS helper functions, so base codefixes will be easier to run on them as well! :)

### DIFF
--- a/packages/migrate/test/__snapshots__/migrate.test.ts.snap
+++ b/packages/migrate/test/__snapshots__/migrate.test.ts.snap
@@ -111,85 +111,7 @@ export default class WithMissingInterface extends Component<WithMissingInterface
 "
 `;
 
-exports[`fix > .gts > component signature codefix > with misssing jsdoc param name 1`] = `
-"import Component from \\"@glimmer/component\\";
-
-export class Something extends Component {
-  /** */
-  // @ts-expect-error @rehearsal TODO TS7019: Rest parameter 'args' implicitly has an 'any[]' type.
-  reset(...args) {
-    console.log(args);
-  }
-
-  shouldExist() {
-    console.log(\\"should exist in output\\");
-  }
-}
-"
-`;
-
 exports[`fix > .gts > component signature codefix > with typedef for component signature interface 1`] = `
-"import Component from \\"@glimmer/component\\";
-
-export interface RepeatSignature {
-  Args: { phrase: any };
-}
-
-//
-// UseCase: Should generate a component signature interface
-//
-
-export class Repeat extends Component<RepeatSignature> {
-  <template>
-    <span>{{@phrase}}</span>
-  </template>
-}
-
-//
-// UseCase: Should infer relationship from comment, update heritage clause with signature
-//
-
-interface MyComponentSignature {
-  Args: { snack: any; age: any };
-}
-
-/**
- * @extends {Component<MyComponentSignature>}
- */
-export class MyComponent extends Component<MyComponentSignature> {
-  name = \\"Bob\\";
-
-  <template>
-    <Repeat @phrase={{@age}} />
-    <span>Hello, I am {{this.name}} and I am {{@age}} years old!</span>
-    <span>My favorite snack is {{@snack}}.</span>
-  </template>
-}
-
-//
-// UseCase: Should generate an interface from all comments
-//
-
-interface InterfaceFromCommentSignature {
-  Args: InterfaceFromCommentArgs;
-}
-
-interface InterfaceFromCommentArgs {
-  someArg: number;
-}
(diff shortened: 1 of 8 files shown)


wondersloth:
This PR should fix #1220

packages/migrate/test/__snapshots__/migrate.test.ts.snap
```diff
@@ -111,23 +111,6 @@ export default class WithMissingInterface extends Component<WithMissingInterface
...
-    console.log(args);
-  }
-
-  shouldExist() {
-    console.log(\\"should exist in output\\");
-  }
-}
-"
-`;
-
```
wondersloth, line 130: These should not be removed? Using test.only

egorio: Can't find test with this name, seems it was renamed... there is same result in `exports[`fix > .gts > strips jsdoc param with missing name 1`] = ``.

Same with the test below.

egorio: image

packages/service/src/glint-service.ts
```diff
@@ -211,7 +243,7 @@ export function severityForDiagnostic(ts: TS, diagnostic: ts.Diagnostic): Diagno
 
 function createSyntheticSourceFile(ts: TS, source: GlintSourceFile): ts.SourceFile {
   return Object.assign(
-    ts.createSourceFile(source.filename, source.contents, ts.ScriptTarget.Latest),
+    ts.createSourceFile(source.filename, source.contents, ts.ScriptTarget.Latest, true),
```
egorio, line 246: This makes SourceFile object similar to those we got from LanguageService with getStart(), getText() etc.

packages/migrate/test/migrate.test.ts
```diff
@@ -971,7 +971,7 @@ export default class SomeComponent extends Component {
     test('previously migrated .gts', async () => {
       await project.write();
 
-      const [inputs, outputs] = prepareInputFiles(project, ['already-migrated.gjs']);
+      const [inputs, outputs] = prepareInputFiles(project, ['already-migrated.gts']);
```
egorio, line 974: :face-palm: there is no `already-migrated.gjs`

packages/service/src/glint-service.ts
```diff
@@ -71,10 +70,12 @@ export class GlintService implements Service {
...
     if (isGlintFile(this, fileName)) {
-      return createSyntheticSourceFile(this.ts, {
-        filename: fileName,
-        contents: this.getFileText(fileName),
-      });
+      return ts.createSourceFile(
+        fileName,
+        this.getFileText(fileName),
+        ts.ScriptTarget.Latest,
+        true
```
egorio, line 77: This `true` makes SourceFile object similar to those we got from LanguageService with getStart(), getText() etc. and now they are compatible with TS utility functions and almost compatible with our own codefixes

packages/service/src/glint-service.ts
```diff
@@ -92,11 +93,39 @@ export class GlintService implements Service {
...
+
+        return undefined;
+      }
+
+      return forEachChild(node, visitEveryNode);
+    };
+
+    visitEveryNode(sourceFile);
+
+    return diagnostics;
```
wondersloth, line 128: Have you considered making an abstract base class with this method in there, so we can share the implementation?

packages/service/src/glint-service.ts
wondersloth: Unsure why we had the synthetic file in the past?

wondersloth (changes requested):

packages/service/src/glint-service.ts
egorio: Just want to keep less changes in this PR. Will "merge" those classes next time once we confirm everything working

packages/service/src/glint-service.ts
egorio: We still have synthetic file, but don't need to populate `text: source.contents` because it's done because the `true` in the 4th param.

wondersloth (commented):
Looks good. Please regress against internal adopter, to verify no new bugs.

lynchbomb (approved):

Issue #1216, opened 2023-08-02
bug: Attempted to transform a range across two different files


lynchbomb:
Current hypothesis is an issue with the @typedef. Task within this ticket is implement the following as a test case in Rehearsal. Thank you @mikrostew for posting:

```
$ rehearsal fix packages/addons/audience-analytics/addon
info:    @rehearsal/fix 1.1.0
✔ Initialize
✔ Analyzing project dependency graph
✖ Attempted to transform a range across two different files 
```

```
// packages/addons/audience-analytics/addon/components/audience-demographics.hbs

{{yield (hash foo="bar")}}
```
``` 
// packages/addons/audience-analytics/addon/components/audience-demographics.ts

import Component from '@glimmer/component';

export default class AudienceDemographics extends Component {
  /**
   * List of items to be rendered in the bar chart
   * @typedef {{ count: number }} HeadCount
   * @returns {HeadCount}
   */
  get something() {
    return {
      count: this.args.someNumber,
    };
  }
}
```

egorio:
Created a ticket with [a related issue](https://github.com/rehearsal-js/rehearsal-js/pull/1226#issuecomment-1668703059) in TypeScript - https://github.com/microsoft/TypeScript/issues/55308

egorio:
It seems the issue "Attempted to transform a range across two different files" is happening when diagnostic.fix plugin is breaking the file syntax and then getDiagnostics() is trying to read that broken content with greek symbols (both should happen under glint service control). 

I feel that the test case above not the only way to get that error. 

Could be wrong.

PR #1226, opened 2023-08-07
test: smoke-test usecase with typedef

lynchbomb:
Ref: https://github.com/rehearsal-js/rehearsal-js/issues/1216

### DIFF
--- a/packages/smoke-test/test/commands/__snapshots__/fix.test.ts.snap
+++ b/packages/smoke-test/test/commands/__snapshots__/fix.test.ts.snap
@@ -500,45 +500,8 @@ exports[`fix command > ember_js_app_4.11 --mode=single-pass 1`] = `
 [DATA] processing file: /app/app.ts
 [DATA] processing file: /app/router.ts
 [DATA] processing file: /app/adapters/application.ts
-[DATA] processing file: /app/components/hello-world.gts
-[DATA] processing file: /app/components/jumbo.hbs
-[DATA] processing file: /app/components/map.hbs
-[DATA] processing file: /app/components/map.ts
-[DATA] processing file: /app/components/nav-bar.hbs
-[DATA] processing file: /app/components/rental.hbs
-[DATA] processing file: /app/components/rentals.hbs
-[DATA] processing file: /app/components/rentals.ts
-[DATA] processing file: /app/components/share-button.hbs
-[DATA] processing file: /app/components/share-button.ts
-[DATA] processing file: /app/models/rental.ts
-[DATA] processing file: /app/routes/index.ts
-[DATA] processing file: /app/routes/rental.ts
-[DATA] processing file: /app/serializers/application.ts
-[DATA] processing file: /app/services/locale.ts
-[DATA] processing file: /app/templates/about.hbs
-[DATA] processing file: /app/templates/application.hbs
-[DATA] processing file: /app/templates/contact.hbs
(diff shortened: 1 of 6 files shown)


wondersloth:
@lynchbomb it's understood this PR is in-progress, just wanted to bring it to your attention.

@egorio mentioned that this PR updated snapshots and is dropping files after the newly added typedef test case.

I pulled down the branch and ran the tests, its was passing, but it seems like we need some assertions for CLI output.

I added something like the following:

```
diff --git a/packages/smoke-test/test/commands/fix.test.ts b/packages/smoke-test/test/commands/fix.test.ts
index e3995233..de660956 100644
--- a/packages/smoke-test/test/commands/fix.test.ts
+++ b/packages/smoke-test/test/commands/fix.test.ts
@@ -69,7 +69,9 @@ describe('fix command', () => {
       }
     );
 
-    expect(cleanOutput(result.stdout, project.baseDir)).toMatchSnapshot();
+    const output = cleanOutput(result.stdout, project.baseDir);
+    expect(output).contains('/app/components/jumbo.hbs');
+    expect(output).toMatchSnapshot();
     expectFile(join(projectRoot, 'app/components/map.hbs')).toMatchSnapshot();
     expectFile(join(projectRoot, 'app/components/map.ts')).toMatchSnapshot();
     expectFile(join(projectRoot, 'app/components/share-button.ts')).toMatchSnapshot();
```

This start to show the test failing.

egorio:
@lynchbomb 

There is a problem with @typedef codefix, I think.

Typescript's typedef's codefix doesn't care about where where the @typedef comment is.
In the example, the @typedef is a part of a method's comment.

```
export default class AudienceDemographics extends Component {
  /**
   * List of items to be rendered in the bar chart
   * @typedef {{ count: number }} HeadCount
   * @returns {HeadCount}
   */
  get something() {
    return {
      count: this.args.someNumber,
    };
  }
}
```

If you would run codefix in IDE (you can test it [there](https://www.typescriptlang.org/play?ssl=10&ssc=2&pln=1&pc=1#code/KYDwDg9gTgLgBAE2AMwIYFcA28DGnUDOBcAguggJbAB2OwAIsALYQDmUqYAFhTsQN4AoOHAD0AKnHCR4uABkKBeBGRwKMZsRgQ4AI2BwoNJEYRrqcGFwO7UUODi52Y0uLIACMAJ5hgSVfz8DhDo1DAAXHDU6Ez69gC+8XAAEsCoCADCIWGuHkYw6FDUAqnpWaEw8bmi0qzA8AQQTPU81KwAFACUcEIihvWFFtGYmADc0lXxQA)), it will just brakes file's syntax because it will adds type in the middle of class:

```
export default class AudienceDemographics extends Component {
  type HeadCount = { count: number; };
  get something(): HeadCount {
    return {
      count: this.args.someNumber,
    };
  }
}
```

Seems we need to improve [@typedef codefix] (https://github.com/egorio/TypeScript/blob/main/src/services/codefixes/convertTypedefToType.ts) to add type in the scopes where they are appropriate to be - file root or current namespace root.

egorio:
@lynchbomb FIY: It fails on `service.getDiagnostics(fileName);` in `DiagnosticCommentPlugin` running under `GlintService` on `.ts` file with "greek" symbols (𝚪, χ).

Not sure where those greek symbols came from in this case, but feel this is somehow related to a broken syntax because of comment above.

egorio:
This case is fixed

PR #1230, opened 2023-08-08
fix: @typedef converted to type in a wrong place

egorio:
Fixes https://github.com/rehearsal-js/rehearsal-js/issues/1229


packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -191,9 +192,26 @@ export class TypescriptCodeFixCollection implements CodeFixCollection {
     return undefined;
   }
 
+  protected makeCodeFixSafe(fix: CodeFixAction, diagnostic: DiagnosticWithContext): CodeFixAction {
+    // Filter type from @typedefs if they are generated outside the root of the source file
+    if (fix.description.startsWith('Convert typedef')) {
```
egorio, line 197: startWith works for both `glint` and `rehearsal` services

packages/codefixes/test/ts-codefixes.test.ts
```diff
@@ -175,6 +175,11 @@ describe('ts-codefixes', () => {
       'convertTypedefToType/failing/fail-2.ts',
       'convertTypedefToType/passing/pass-2.ts'
     );
+
+    await runTest(
+      'convertTypedefToType/failing/fail-3.ts',
+      'convertTypedefToType/passing/pass-3.ts'
```
wondersloth, line 181: How do these tests differ?

packages/smoke-test/test/commands/__snapshots__/fix.test.ts.snap
```diff
@@ -723,6 +725,27 @@ export default class HelloWorld extends Component<HelloWorldCommentSignature> {
...
+"import Component from '@glimmer/component';
+
+export default class AudienceDemographics extends Component {
+  /**
+   * List of items to be rendered in the bar chart
+   * @typedef {{ count: number }} HeadCount
+   * @returns {HeadCount}
+   */
+  // @ts-expect-error @rehearsal TODO TS2304: Cannot find name 'HeadCount'.
+  get something(): HeadCount {
```
wondersloth, line 740: Is this the expected result? We should not create the interface because it won't be created safely?

packages/ts-utils/src/helpers/typescript-ast.ts
```diff
@@ -46,15 +46,23 @@ export function isTheFirstParentNodeInTheLIne(node: Node): boolean {
...
+  includeJSDocNodes = false
 ): Node | undefined {
   const visitor = (node: Node): Node | undefined => {
     if (isNodeAtPosition(node, start, length)) {
       return node;
     }
 
-    if (node.getStart() <= start && node.getEnd() >= start + length) {
-      return ts.forEachChild(node, visitor);
+    if (includeJSDocNodes) {
```
wondersloth, line 57: Isn't the JSDocNodes dependent on how the file is parsed?

packages/codefixes/test/ts-codefixes.test.ts
egorio: The `fail-3.ts` is incoming file, the `pass-3.ts` is output ("snapshot"). We don't expect any changes in cases like this, this is why they are the same.

packages/smoke-test/test/commands/__snapshots__/fix.test.ts.snap
egorio: For now we decided do not create type if it's not it the right place (if created type not in the SourceFile) to unblock migration. The next idea is to, probably, modify codefix result to put type in the root / namespace / function body.

+ https://github.com/microsoft/TypeScript/issues/55308

packages/ts-utils/src/helpers/typescript-ast.ts
egorio: Unfortunately no, they are always not a part of a normal AST tree (node.+children) but instead part of `node.JSDocs`.

packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -202,6 +220,10 @@ export class TypescriptCodeFixCollection implements CodeFixCollection {
         continue;
       }
 
+      if (filter.safeFixes) {
+        fix = this.makeCodeFixSafe(fix, diagnostic);
```
mikesherov, line 224: would it better for this function to return a boolean, and `continue` on `false`, rather than modifying the `changes` property and doing a check for empty changes?

egorio: It's more flexible to modify changes (we do same in makeCodeFixStrict).

In the next step it expected to return modified changes with 2 items: remove comment + add type somewhere in the file.

Plus codefixe can have few changes and only one can be modified / removed in case we will be using this function for something else in future.

mikesherov: in that case, should the 

```
// Covers cases with broken type signatures, like: `() =>`
        const brokenTypeSignatures = /\(\) =>\s*$/i;
        if (brokenTypeSignatures.test(textChanges.newText)) {
          continue;
 }
```

code from strict move here?

egorio: Type signature provided by codefix isn't broken `type HeadCount = { count: number; };`

But, file became broken after it's added to the file (codefix applied) at the wrong place.

egorio: Also, we do check for broken signatures / unresolvable types in `makeCodeStrict`

egorio: Also, we do check for broken signatures / unresolvable types in `annotateWithStrictTypeFromJSDoc` since JSDocs are the source of broken types :)

lynchbomb (approved):

Issue #1228, opened 2023-08-07
feat(glint): add support for stubbing missing arguments for usage of this.args in component class


wondersloth:
### Summary

Imp[rove `AddMissingArgToComponentSignature` to support populating the ComponentSignature this usages of a `this.args.someArg`.

```typescript
import Component from \\"@glimmer/component\\";;

export interface SalutationComponentSignature {
  Args: {
    keywords: any;
  }
}

export default class Salutation extends Component<SalutationComponentSignature> {
  get fancyName() {
    // @ts-expect-error @rehearsal TODO TS2339: Property 'name' does not exist on type 'Readonly<{ keywords: any; }>'.
    return `Super Duper ${this.args.name}`;
   }
  <template>
    {{@keywords}}
    {{this.fancyName}}
  </template>
}
```

Current logic is bound for inferring that we're in a template region of a gts file and inferring the arugment name being used.

This feature still operates on TS2339 codes. 

### Recommended Changes

- [ ] Refactor `addMissingArgToComponentSignature` https://github.com/rehearsal-js/rehearsal-js/blob/52a58a283045439501a5f65212b9e85374a36fac/packages/codefixes/src/fixes/glint/addMissingArgToComponentSignature.ts#L55 to check if it's in a context where it's invoking `this.args.someArg`

PR #1232, opened 2023-08-13
feat(glint): add this.args usage to signature

wondersloth:
This addresses #1228, Adds support for populating component signatures with usages like: `this.args.someParam` in `gts` files.


egorio (approved):
Looks good!

PR #1236, opened 2023-08-15
fix: prettier.format() return empty string for some .gts files

egorio:
Prettier 3.0.0 doesn't fail on .gts files.

This solution can unblock users, but we probably need to think about using the project's version of prettier (and eslint?) to format the project's content (probably even run bin files).


wondersloth (changes requested):
Update the README.md

package.json
```diff
@@ -57,7 +57,7 @@
     "@types/lodash.get": "^4.4.7",
     "@types/micromatch": "^4.0.2",
     "@types/node": "^20.3.3",
-    "@types/prettier": "^2.7.2",
+    "@types/prettier": "^3.0.0",
```
wondersloth, line 60: We don't need the `@types/prettier` https://prettier.io/blog/2023/07/05/3.0.0.html#api-2

wondersloth:
Also, the glimmerx prettier plugin will no longer work in prettier 3.0. So we need to document that it needs to be updated.

wondersloth:
There are still a lot of fixtures and test-cases that use `2.*` dependencies. We need to ensure we have back compat on any update or update all instances.

wondersloth (changes requested):

lynchbomb:
Lets just:
- bump prettier as a rehearsal dep to ^3
- bump prettier pre-flight check as hard dev-dep to ^3 here: (https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/cli/src/prereqs.ts#L39)
- we might _also_ want to bump typescript in the pre-flight check to ^5 with this change
- after this lands we will release a new major

Issue #1235, opened 2023-08-15
Rehearsal removes a content of .gts files after prettier.format()


egorio:
Rehearsal sometimes empties .gts files because prettier.format(text, prettierOptions); returns an empty string.

PR #1237, opened 2023-08-15
fix: prettier.format() return empty string for some .gts files

egorio:
Upgrade `prettier` to `3.0.2` across packages and test fixtures.

Pre-request `prettier` 3.0.0+ and `prettier-plugin-ember-template-tag` 1.0.0+ to be installed.

The PrettierPlugin migrated to async functions.

Diagnostics-fix and Glint-fix plugins now calculate format code settings for the processing file and pass them down to CodeFix collection (`CodeFixCollection.getFixesForDiagnostic`, `CodeFix.getCodeAction`) similar to TS's `LanguageService.getCodeFixesAtPosition`). Not all code fixes can use those settings to produce pre-formatted code changes.

Fixes: https://github.com/rehearsal-js/rehearsal-js/issues/1235


packages/codefixes/src/typescript-codefix-collection.ts
```diff
@@ -86,45 +80,6 @@ export class TypescriptCodeFixCollection implements CodeFixCollection {
     };
   }
 
-  protected getFormatCodeSettingsForFile(filePath: string): FormatCodeSettings {
```
egorio, line 89: Moved to helper to be called in fix plugins and pass settings to getCodefix* funstions

egorio: This helps to deal with an async version of the prettier + function will be called once per file + all getCodefix functions can utilize settings when generating code changes.

lynchbomb (approved):

PR #1242, opened 2023-08-18
feat: add return types to functions

wondersloth:
Paired with @egorio
- Overall improves return types for function/methods.
- Renames `addMissingTypesBasedOnInlayHints` to `addMissingReturnTypes`
- Infers return types from signatures using typescript API. 
- Fallback to inlay hints when above fails.
- Refactor finding closing paren logic across codefix plugins and move to `packages/tsc-utils`.

TODO
- [x] Investigate the try/catch failures where the types were not inferred.
  - In `.gts` files where a function or method is using spread operator on arguments, getting the component signature will fail. We added the try/catch to work around these use cases. Added back in the inlay hints as a fallback so we can get some types. Ideally we should be to construct these return signatures in some way as we can see them in the IDE Hover overlays.


.npmrc
```diff
@@ -1 +1 @@
-use-node-version=16.19.0
+registry=https://registry.npmjs.org/
```
wondersloth, line 1: Remove this, I needed this on my machine, because it was attempting to fetch from artifactory for some reason.

packages/migrate/test/__snapshots__/migrate.test.ts.snap
```diff
@@ -213,8 +214,8 @@ export interface SomethingSignature {
 
 export class Something extends Component<SomethingSignature> {
   /** */
-  // @ts-expect-error @rehearsal TODO TS7019: Rest parameter 'args' implicitly has an 'any[]' type.
-  reset(...args): void {
+  // @ts-expect-error @rehearsal TODO TS7050: Function 'reset' lacks a return-type annotation.
+  reset(...args) {
```
wondersloth, line 218: This isn't ideal, because we had `void`

The `...args` fails in `.gts` file because of a missing `symbol` on the parameter node. 

We could backstop this with bring back our inlay hints logic when this fails the `try..catch`

wondersloth (commented):
- [x] Investigate the try/catch failures where the types were not inferred.

packages/codefixes/src/fixes/addMissingReturnTypes.ts
```diff
@@ -1,32 +1,75 @@
...
+        return;
+      }
+
+      typeToString = diagnostic.checker.typeToString(returnType);
+    } catch (error) {
+      // Fallback to inlay hints if unable to determine return type of signature
+      return this.fixMissingReturnTypeWithInlayHints(diagnostic, targetPosition);
+    }
+
+    if (!typeToString || typeToString.includes('any') || typeToString.includes('object')) {
```
egorio, line 51: Probably need to return check for '...' for a "Fallback to inlay hints"

egorio (approved):

Issue #1246, opened 2023-08-22
bug: rehearsal fix adds empty lines between comments in subsequent runs


egorio:
If run `rehearsal fix` multiple times on:

```
export class Foo {
  hello() {
    // Existing comment
    return this.name;
  }
}
```

The first pass will result:
```
export class Foo {
  hello() {
    // Existing comment
    // @ts-expect-error @rehearsal TODO TS2339: Property 'name' does not exist on type 'Foo'.
    return this.name;
  }
}
```

but the second and following runs will add empty lines between an existing comment and `@ts-expect-error` comment:

```
export class Foo {
  hello() {
    // Existing comment

    // @ts-expect-error @rehearsal TODO TS2339: Property 'name' does not exist on type 'Foo'.
    return this.name;
  }
}
```

Expected Rehersal to output similar results without adding empty comments

PR #1247, opened 2023-08-22
fix: do not add empty lines before @ts-expect-error on sequential runs

egorio:
The `re-rehearse` plugin removes existing comments with `@rehearsal` without including `\n`.

This PR fixes the issue and `@rehearsal` comments will be removed correctly (including `\n` at the end).

Fixes https://github.com/rehearsal-js/rehearsal-js/issues/1246

### DIFF
--- a/packages/plugins/src/plugins/diagnostic-report.plugin.ts
+++ b/packages/plugins/src/plugins/diagnostic-report.plugin.ts
@@ -4,8 +4,7 @@ import { findNodeAtPosition } from '@rehearsal/ts-utils';
 import debug from 'debug';
 import ts from 'typescript';
 import { getLocation } from '../helpers.js';
-
-const { isLineBreak } = ts;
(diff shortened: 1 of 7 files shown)


packages/plugins/src/plugins/diagnostic-report.plugin.ts
```diff
@@ -20,10 +19,10 @@ export class DiagnosticReportPlugin extends Plugin<DiagnosticReportPluginOptions
 
     DEBUG_CALLBACK(`Plugin 'DiagnosticReport' run on %O:`, fileName);
 
-    const originalConentWithErrorsSuppressed = context.service.getFileText(fileName);
+    const originalContentWithErrorsSuppressed = context.service.getFileText(fileName);
```
egorio, line 22: Fixed few typos in this file

lynchbomb (approved):

PR #1249, opened 2023-08-23
feat: add strict type from jsdocs to .gts files

wondersloth:
Co-authored-by: @egorio 

- Added `AnnotateWithStrictTypeFromJSDoc` to the list of `glintCodeFixes`.
- Refactored `AnnotateWithStrictTypeFromJSDoc` to support a glint diagnotics.
  - This is not ideal, but it works. Ideally we shouldn't muddle an existing CodeFix for TS with glint specifics.
- Removed  diagnostic `95009` from `annotateWithStrictTypes.messages` in `packages/codefixes/src/codefixesMessages.json`.
```
  "annotateWithStrictTypeFromJSDoc": {
    "title": "Annotate Strict Types From JSDoc",
    "description": "Annotates code with the only strict types from the declared JSDoc comments. Strict replacement for annotateWithTypeFromJSDoc",
    "diagnostics": [80004],
-    "messages": [95009],
+    "messages": [],
    "builtIn": true
  },
```
This diagnostic message was causing the our strict types fix to be filtered out.

### DIFF
--- a/packages/codefixes/src/fixes/annotateWithStrictTypeFromJSDoc.ts
+++ b/packages/codefixes/src/fixes/annotateWithStrictTypeFromJSDoc.ts
@@ -1,9 +1,11 @@
 import { canTypeBeResolved, findNodeEndsAtPosition } from '@rehearsal/ts-utils';
 import ts, { FormatCodeSettings } from 'typescript';
 import { Diagnostics } from '../diagnosticInformationMap.generated.js';
+// import { TypescriptCodeFixCollection } from '../typescript-codefix-collection.js';
 import { TypescriptCodeFixCollection } from '../typescript-codefix-collection.js';
+import { GlintCodeFixCollection, GlintDiagnosticWithContext } from '../glint-codefix-collection.js';
 import type { FileTextChanges, TextChange, CodeFixAction } from 'typescript';
-import type { CodeFix, DiagnosticWithContext } from '../types.js';
+import type { CodeFix, CodeFixCollection, DiagnosticWithContext } from '../types.js';
 
 const {
   isFunctionDeclaration,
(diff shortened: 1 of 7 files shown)


packages/codefixes/src/fixes/annotateWithStrictTypeFromJSDoc.ts
```diff
@@ -16,16 +18,31 @@ const {
...
+  diagnostic: unknown
+): diagnostic is GlintDiagnosticWithContext {
+  return !!diagnostic && Object.hasOwnProperty.call(diagnostic, 'glintService');
+}
+
 export class AnnotateWithStrictTypeFromJSDoc implements CodeFix {
   getErrorCodes = (): number[] => [Diagnostics.TS80004.code];
 
+  getCodeFixCollection(diagnostic: DiagnosticWithContext): CodeFixCollection {
+    return isGlintDiagnosticWithContext(diagnostic)
```
egorio, line 31: 🥇 This does the trick to get fixes at the right positions for .gts!

egorio (approved):

packages/migrate/test/migrate.test.ts
```diff
@@ -147,6 +147,29 @@ describe('fix', () => {
...
+        filesToMigrate: inputs,
+        reporter,
+        mode: 'drain',
+      };
+
+      for await (const _ of migrate(input)) {
+        // no ops
+      }
+
+      expectFile(outputs[0]).contains('export const a: number | string = 0;');
```
lynchbomb, line 166: Is this just a backup to the snapshot?

lynchbomb (approved):

packages/migrate/test/migrate.test.ts
wondersloth: Yes. It's one easier to i debug as the assertion is succinct. Secondly it helps prevent in those cases a snapshot is updated with bad data.

PR #1271, opened 2023-09-05
chore: unwind circular monorepo deps

lynchbomb:
This PR resolves circular deps we had with monorepo imports. 

- moves the Plugin class into the plugins package `packages/service/src/plugin.ts` > `packages/plugins/src/plugin.ts`
- swaps deps for dev-deps where needed
- updates imports based on new location of plugin class

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,8 +82,8 @@
     "tmp": "^0.2.1",
     "ts-node": "^10.9.1",
     "typescript": "^5.2.2",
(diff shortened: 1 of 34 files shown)


wondersloth (approved):
Has some test failures that need investigation, otherwise looks good.

egorio (approved):

wondersloth (approved):

Issue #1227, opened 2023-08-07
feat(glint): support creating component signature for template only components .gts files.


wondersloth:
Current support is for only .gts files with component backing classes. We should support template only components in a .gts files.

This feature should support:
- [X] Adding an argument to an existing signature. Addressed by #1276 
- [X] Adding an `Args` property to an empty signature.  Addressed by #1276 
- [X] Generate a complete signature if none exists. #1278
- [X] Add support for type `TOC` <-- Must Have, Necessary for internal adopters of TemplateOnlyComponents #1278 
- [ ] Add support for renamed type <-- Should Have, This doesn't happen often but could happen.
- [ ] Add support for missing default export <-- Could Have, Need to evaluate this use case.

PR #1274, opened 2023-09-15
feat(codefix): add parse utils for template only components

wondersloth:
- `getNearestTemplateOnlyComponentVariableDeclaration` will be used to walk up from a glint args diagnostic to a template only declaration.
- `getComponentSignatureNameFromTemplateOnlyComponent` will be used to grab an existing signature for a TOC.
- Supports feature work for  #1227 
- Refactor getNearest methods to use `ts.findAncestor` library method


packages/codefixes/src/fixes/glint/glint-parsing-utils.ts
```diff
@@ -243,3 +243,57 @@ export function getJSDocExtendsTagWithSignature(
...
+  ) {
+    return true;
+  }
+
+  return false;
+}
+
+export function getNearestTemplateOnlyComponentVariableDeclaration(
+  targetNode: ts.Node,
+  identifierName = 'TemplateOnlyComponent'
```
egorio, line 270: Can this value be a constant?

egorio (approved):

packages/codefixes/src/fixes/glint/glint-parsing-utils.ts
```diff
@@ -243,3 +243,57 @@ export function getJSDocExtendsTagWithSignature(
...
+  let target = targetNode;
+  do {
+    if (!target.parent) {
+      return;
+    }
+
+    if (isTemplateOnlyComponent(target, identifierName)) {
+      return target;
+    }
+  } while ((target = target.parent));
```
egorio, line 281: You might want to try ts.findAncestor, like this: https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/codefixes/src/fixes/addMissingTypesBasedOnInheritance.ts#L65

wondersloth: Thank you for the recommendation. Will update this to use the library method.

wondersloth: I've pushed up  changes to the two methods that were finding up.

egorio (approved):

PR #1304, opened 2023-10-30
(fix) Updating glint-service to use the exported `URI` method from vscode-uri

robostheimer:
## What
Currently when any rehearsal command is used, the following error is emitted:
```       ^^^^^^^^^
SyntaxError: The requested module 'vscode-uri' does not provide an export named 'default'
```

There appears to have been api change in vscode-uri that rehearsal picked up as part of a package update.  This diff fixes the issue by using the exported URI function from vscode-uri.

Note, error will only occur if consuming app is on vscode-uri 3.0.8.  However, we should keep up with the package upgrade and ensure that no consumers are blocked by the above issue.

Note, we also need to bump glint/core dependency which contains the same issue being fixed in this diff: https://github.com/typed-ember/glint/blame/main/packages/core/src/language-server/util/index.ts

## Testing
- `pnmp vitest -u` passes when run on `packages/service` directory
```
✓ test/rehearsal-service.test.ts (6) 2430ms
   ✓ Test service (6) 2430ms
     ✓ construct
     ✓ getFileText
     ✓ setFileText
     ✓ setFileText
     ✓ getSourceFile 1542ms
     ✓ getDiagnostics 878ms

 Test Files  1 passed (1)
      Tests  6 passed (6)
   Start at  09:06:31
   Duration  6.79s (transform 3.06s, setup 0ms, collect 3.77s, tests 2.43s, environment 0ms, prepare 204ms)
```
- manual testing shows that the fix works and rehearsal commands can be run successfully with this diff.

### DIFF
--- a/packages/plugins/package.json
+++ b/packages/plugins/package.json
@@ -58,7 +58,7 @@
     "winston": "^3.8.2"
   },
   "peerDependencies": {
-    "@glint/core": "^1.1.0",
+    "@glint/core": "^1.2.1",
     "typescript": "^5"
   },
   "packageManager": "pnpm@8.6.7",
--- a/packages/service/src/glint-service.ts
+++ b/packages/service/src/glint-service.ts
@@ -2,7 +2,7 @@
 // about how it should be imported (╯°□°)╯︵ ┻━┻
 // eslint-disable-next-line import/default
 import { URI } from 'vscode-uri';
-import * as ts from 'typescript';
+import ts from 'typescript';
 import {
   DiagnosticSeverity,
   type TextDocumentEdit,
--- a/packages/service/src/glint-service.ts
+++ b/packages/service/src/glint-service.ts
@@ -1,8 +1,8 @@
 // This is disabled because vscode-uri is a commonjs module, which causes TS and Eslint to disagree
 // about how it should be imported (╯°□°)╯︵ ┻━┻
 // eslint-disable-next-line import/default
-import VSCodeURI from 'vscode-uri';
-import ts from 'typescript';
+import { URI } from 'vscode-uri';
+import * as ts from 'typescript';
 import {
   DiagnosticSeverity,
   type TextDocumentEdit,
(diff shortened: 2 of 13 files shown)


packages/service/src/glint-service.ts
```diff
@@ -1,7 +1,7 @@
 // This is disabled because vscode-uri is a commonjs module, which causes TS and Eslint to disagree
 // about how it should be imported (╯°□°)╯︵ ┻━┻
 // eslint-disable-next-line import/default
```
mikesherov, line 3: Is this disabling comment needed anymore? I think that's specifically what 3.0.8 fixes.

robostheimer: You are correct.  Will remove.

mikesherov (approved):
LGTM, will wait for @rwjblue to review before merging. Thank you!

packages/plugins/package.json
```diff
@@ -58,7 +58,7 @@
     "winston": "^3.8.2"
   },
   "peerDependencies": {
-    "@glint/core": "^1.1.0",
+    "@glint/core": "^1.2.1",
```
rwjblue, line 61: Generally speaking bumping the minimum of a package in peer dependencies is a breaking change. However, for this specific case I searched GitHub (OSS) and internal repos for reference to `@rehearsal/plugins` (outside of lockfiles), and only found `@rehearsal/migrate` (which is also updated in this PR). So this is practically speaking going to be fine.

rwjblue (approved):

packages/plugins/package.json
rwjblue, line 61: Generally speaking bumping the minimum of a package in peer dependencies is a breaking change. However, for this specific case I searched GitHub (OSS) and internal repos for reference to @rehearsal/plugins (outside of lockfiles), and only found @rehearsal/migrate (which is also updated in this PR). So this is practically speaking going to be fine.

robostheimer: Ah, gotcha.  I was actually surprised that vscode-uri's change wasn't actually a major change.  They changed the api which caused consuming apps like rehearsal to mis-behave.

